PR #1183, opened 2014-06-07
abstract-php: Enable building embed target

dholm:
Adds an option to build the PHP embed target (PHP as a library).  This
feature can be used with uWSGI, for instance.


alanthing:
I tried to build with and without the expression change mentioned by @lucasmichot and had compiler errors each time. Can you rebase with master (since we're several months from the original request) and push your branch again so the test fires and we can all see what the Brew Bot output is?

alanthing:
I'm going to close this PR as it needs a lot of refactoring to account for recent changes. A new PR should be submitted to add embed options.

PR #1328, opened 2014-09-15
Added formula for phpMyAdmin 3

szinya:
For those who like the 3.x interface better.


alanthing:
I made a new formula based on *phpmyadmin.rb* that would not conflict and allow both to live side-by-side. Also, this PR had a commit that wanted to remove hundreds of files. 

Since those electing to choose version 3.x will be in the minority, I elected not to rename the existing phpMyAdmin formula to *phpmyadmin4.rb* and instead point to a "default" without a number. I could see how this may lead some to think "there's a formula without a number and another with a '3', the '3' must be a higher value and therefore newer," so I'm certainly willing to be overridden on that decision.

PR #1340, opened 2014-09-24
update runkit

denji:


alanthing:
Closing since this was implemented in b6f792c. If it was remaining open for the upstream maintainer to tag a release, I doubt it'll happen since it hasn't happen in nearly 5 years :frowning:

PR #1368, opened 2014-10-10
Add enchant http://php.net/manual/en/book.enchant.php

calvintennant:


alanthing:
I refactored to make it optional. Thanks for the contribution!

PR #1454, opened 2014-11-27
Upgrade mysqlnd_ms version to 1.5.2

lkebin:


alanthing:
Closing since upstream is now up to 1.6.0 and another PR should be submitted for the newer version if this one is not going to be fixed with a patch.

PR #1473, opened 2014-12-08
Enable opcache after install

romainneutron:
PHP 5.5 and 5.6 are compiled with opcache. However, this one is not enabled. This PR solves this


alanthing:
I refactored this to address a couple of concerns. 

First, no resulting build of upstream (PHP itself) will result in `opcache.enabled=1` being set by default. This is good, because the default `opcache.memory_consumption` is a meager *64*, which could cause problems on local development with anything remotely complicated if the available memory ran out. 

Secondly, we're set up use upstream's [php.ini-development](https://github.com/php/php-src/blob/master/php.ini-development) as our initial php.ini file. It's also not enabled there, nor by any PHP install scripts. However, most operating system distributions will use another method to at least include the `opcache.so` extension in a way that fits whatever method deemed appropriate by the OS maintainers (for example, Red Hat uses "/etc/php.d/\*.conf"). While you could argue that the "Homebrew method" is using the *$(brew --prefix)/etc/php/5.6/conf.d* directory, I would say that is for *additional* PHP extensions that are not part of PHP-core, such as PhpRedis.

The end result here is that, since *--enable-opcache* is a default build flag when building PHP with Homebrew, the *opcache.so* extension file is loaded by default like all of the other default extensions. It is still up to the user to configure Opcache manually in php.ini, but I would rather it be in php.ini because Opcache is bundled with PHP rather than an additional/third-party extension. Let me know if my explanation needs additional clarification. And thanks for the PR.

alanthing:
Also, the changes to *php.ini* are only made on the first installation of a `php55` or `php56` formula. For example, you can customize Opcache in *$(brew --prefix)/etc/php/5.6/php.ini* and the changes will not be lost upon an upgrade of `php5x`.

guoxiao:
My pull request #1508 won't build in Jenkins. It seems relative to this #1473 
```
Error: inreplace failed
/usr/local/etc/php/5.4/php.ini:
  expected replacement of /^(\[opcache\].*)$/ with "\\1\n; Load the opcache extension\nzend_extension=opcache.so\n"
  expected replacement of /^;?opcache\.enable\s*=.+$/ with "opcache.enable=0"
```

alanthing:
Acknowledged. Fix coming shortly.

PR #1476, opened 2014-12-17
php53-opcache: Add opcache extension with upstream defaults

alanthing:
ZendOpcache works great with `php53`. I have been installing it in Homebrew via `pecl install zendopcache-beta`, but this will make it a lot easier. It works exactly as it does with `php54`. While not on OS X, I've used the same version opcache with PHP 5.3 via pecl in production for several months with great success.

Instead of the minimal and non-default configuration from the `php54-opcache.rb` formula, I believe it's more appropriate to use the sample ini with defaults from the upstream project.


alanthing:
This is a new and non-conflicting formula; anything I can do to help get this merged/accepted?

PR #1477, opened 2014-12-17
php54-opcache: Use upstream config defaults

alanthing:
The defaults in `php54-opcache.rb` do not contain comments, and supplied values deviate from the upstream defaults. Particularly, the higher value for `opcache.revalidate_freq` of a high value of 60 seconds can introduce confusion in local development. To remove ambiguity, I believe it makes sense to use the ini defaults from the upstream project.


alanthing:
This is a pretty simple change, anything I can do to help get this merged/accepted?

PR #1478, opened 2014-12-17
abstract-php: `pear config-set` commands to use "system" layer

alanthing:
After a version of PHP is installed with `brew`, the following command is run to instruct `pear` where the main `php.ini` is on the filesystem:

    pear config-set php_ini $(brew --prefix)/etc/php/5.x/php.ini

This change is applied to `~/.pearrc` since the default layer for `pear config-set` is "user". This could be problematic if you install multiple versions of PHP and switch between them and you may end up with a "php_ini" value in `~/.pearrc` that is different that the version of php currently linked with `brew`; then, using `pear` or `pecl` to install something could put the package in an unintended location.

The following change will put the value of "php_ini" in `$(brew --prefix)/etc/php/5.x/pear.conf` where it should be by specifying the "system" layer:

    pear config-set php_ini $(brew --prefix)/etc/php/5.x/php.ini system

Because `pear` and `pecl` are compiled to use a static path for the pear.conf file and it's matching up with the related version of PHP's php.ini location, this is the appropriate method and location for this `config-set` value.


alanthing:
This is a pretty simple change, anything I can do to help get this merged/accepted?

PR #1482, opened 2014-12-22
Added 'Require all granted' to httpd.conf example in phpMyAdmin

NathanJang:
When following the original httpd.conf example, `http://localhost/phpmyadmin/` had a 403 error. I followed @james2doyle's recommendation in [issue 1234](https://github.com/Homebrew/homebrew-php/issues/1234) and the error stopped.

If there is a better fix for this, kindly advise.


alanthing:
I refactored this a bit to allow for a single configuration for both Apache 2.2 and 2.4 via `<IfModule mod_authz_core.c>` and `<IfModule !mod_authz_core.c>`. Thanks for the contribution!

NathanJang:
Thanks!

PR #1488, opened 2014-12-31
Use opt_prefix for php-fpm plist paths

gapple:
`opt_prefix` provides a stable path for the installed version of a formula, and is specifically recommended for use in plists (see [formula.rb#L311](https://github.com/Homebrew/homebrew/blob/f4cee536746ae8cd78200b48a11d877d4065c130/Library/Homebrew/formula.rb#L311))


alanthing:
Thanks!

Issue #1494, opened 2015-01-06
failed to install php56-vld


CodeTheWorld:
brew install php56-vld
==> Installing php56-vld from homebrew/homebrew-php
==> Downloading http://pecl.php.net/get/vld-0.12.0.tgz
Already downloaded: /Library/Caches/Homebrew/php56-vld-0.12.0.tgz
==> Patching
==> /usr/local/opt/php56/bin/phpize
==> ./configure --prefix=/usr/local/Cellar/php56-vld/0.12.0 --with-php-config=/usr/local/opt/php56/bin/php-config
==> make
                case IS_CONSTANT_ARRAY: return vld_dump_zval_constant_array (val.value);
                     ^
1 error generated.
make: *** [srm_oparray.lo] Error 1
make: *** Waiting for unfinished jobs....

READ THIS: http://git.io/brew-troubleshooting
If reporting this issue please do so at (not Homebrew/homebrew):
  https://github.com/homebrew/homebrew-php/issues

marcaube:
Got the same error, here's my gist-logs for `php56-vld` https://gist.github.com/ac98dc9fe25396eeee28

PR #1495, opened 2015-01-07
Upgrade VLD extension, fixing #1494

tmoitie:
Upgraded VLD to latest version on PECL - 0.13. Replicated #1494 & version bump appears to fix it.


alanthing:
Thanks!

Issue #1498, opened 2015-01-10
brew install php56-yaf conflict with MacPort 


WangWenchao:
As I have installed the MacPorts , but I use the brew to manange  php dev environment from homebrew-php

but when i install php56-yaf ,get the errors:
 brew install php56-yaf
==> Installing php56-yaf from homebrew/homebrew-php
==> Downloading http://pecl.php.net/get/yaf-2.3.2.tgz
Already downloaded: /Library/Caches/Homebrew/php56-yaf-2.3.2.tgz
==> Patching
==> /usr/local/opt/php56/bin/phpize
==> ./configure --prefix=/usr/local/Cellar/php56-yaf/2.3.2 --with-php-config=/usr/local/opt/php56/bin/php-config
==> make
                     ^
3 errors generated.
make: *** [yaf_config.lo] Error 1
make: *** Waiting for unfinished jobs....
 clang -I. -I/private/tmp/php56-yaf-UtrWbi/yaf-2.3.2 -DPHP_ATOM_INC -I/private/tmp/php56-yaf-UtrWbi/yaf-2.3.2/include -I/private/tmp/php56-yaf-UtrWbi/yaf-2.3.2/main -I/private/tmp/php56-yaf-UtrWbi/yaf-2.3.2 -I/usr/local/Cellar/php56/5.6.4/include/php -I/usr/local/Cellar/php56/5.6.4/include/php/main -I/usr/local/Cellar/php56/5.6.4/include/php/TSRM -I/usr/local/Cellar/php56/5.6.4/include/php/Zend -I/usr/local/Cellar/php56/5.6.4/include/php/ext -I/usr/local/Cellar/php56/5.6.4/include/php/ext/date/lib -DHAVE_CONFIG_H -g -O2 -c /private/tmp/php56-yaf-UtrWbi/yaf-2.3.2/configs/yaf_config_simple.c  -fno-common -DPIC -o configs/.libs/yaf_config_simple.o

**Warning: It appears you have MacPorts or Fink installed.**
Software installed with other package managers causes known problems for
Homebrew. If a formula fails to build, uninstall MacPorts/Fink and try again.

READ THIS: http://git.io/brew-troubleshooting
If reporting this issue please do so at (not Homebrew/homebrew):
  https://github.com/homebrew/homebrew-php/issues

WangWenchao:
I dont know whats that ,and after uninstall the Macport ,get 
brew install php56-yaf
[...]

PR #1496, opened 2015-01-07
Update yaf extension to 2.3.3 to fix #1487

tmoitie:
Replicated issue in ticket #1487 and was able to compile with the latest yaf source.

PR #1499, opened 2015-01-10
Update php56-yaf.rb

WangWenchao:
update to  resolve the compiled error https://github.com/Homebrew/homebrew-php/issues/1498


alanthing:
Fixed in b9293e1 via #1496.

PR #1503, opened 2015-01-12
Depend on homebrew/dupes/zlib, not zlib

gabebw:
Without this change:

  $ brew tap homebrew/dupes
  $ brew install zlib
  $ brew install php56
  ==> Installing php56 from homebrew/homebrew-php
  ==> Downloading https://www.php.net/get/php-5.6.4.tar.bz2/from/this/mirror
  Already downloaded: /Library/Caches/Homebrew/php56-5.6.4
  ==> Patching
  Error: No available formula for dupes/zlib
  Searching formulae...
  Searching taps...

And then it stops.

With this change, it all works.

(I have not tested it with any other PHP formulae.)

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -180,7 +180,7 @@ INFO
       "--enable-mbregex",
       "--enable-bcmath",
       "--enable-calendar",
-      "--with-zlib=#{Formula['zlib'].opt_prefix}",
+      "--with-zlib=#{Formula['homebrew/dupes/zlib'].opt_prefix}",
       "--with-ldap",
       "--with-ldap-sasl=/usr",
       "--with-xmlrpc",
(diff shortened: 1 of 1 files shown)


gabebw:
Hmm, it works on another computer. Odd.

alanthing:
I'm going to close without accepting. We have similar code in [httpd22](https://github.com/Homebrew/homebrew-apache/blob/master/httpd22.rb#L20) and [httpd24](https://github.com/Homebrew/homebrew-apache/blob/master/httpd24.rb#L20) in [homebrew/apache](https://github.com/Homebrew/homebrew-apache). It works fine with `depends_on` using the tap-like path and then referencing `Formula['zlib'].opt_prefix` from my experience. 

You might be experiencing a problem that I've seen in that if you are installing a formula that references a tap that you don't already have tapped, it will error the first time, but proceed successfully the second time. If you run into this problem again with zlib on a fresh system, do one of two things: run `brew tap homebrew/dupes` ahead of time, or simply install the formula again. The first run will do the tap so the second run should proceed successfully.

gabebw:
Thank you for looking into this.

alanthing:
You're welcome! I'll run this issue (referencing a non-tapped formula requires installing twice or tapping ahead of time) up the chain and see what the deal is.

PR #1508, opened 2015-01-18
php-magickwand 1.0.9-2 (new formula)

guoxiao:


guoxiao:
Autobuild is failed? It seems that php is not installed properly?

guoxiao:
The build failure may be relative with pull request #1473

alanthing:
Resolved. Please re-push for another build attempt.

guoxiao:
@alanthing Merged build failed again.
```
Error Message

failed: brew doctor
Stacktrace

        Warning: Some installed formula are missing dependencies.
You should `brew install` the missing dependencies:

    brew install metis parmetis

Run `brew missing` for more details.
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!
```

alanthing:
Looks like a problem with the build server. But, I tried to build this locally and was told that `MagickWand-config` was not in the "default path," despite it actually being in my $PATH. It reported the same on a manual `phpize && ./configure` without Homebrew. Also, I have Homebrew in a custom location (not "/usr/local"). Do you know if this extension has trouble building if ImageMagick is not in /usr/local?

alanthing:
I think that's a build bot problem.

I added `"--with-magickwand=#{Formula['imagemagick'].opt_prefix}",` to the formulas so it looks for ImageMagick in Homebrew (since it's a hard dependency) instead of the system. It failed to build on my system until I added it because I don't have ImageMagick aside from the version it grabbed from Homebrew as a dependency.

Thanks for the contribution!

PR #1509, opened 2015-01-18
added pimple extension formula (53-56)

noodlehaus:
This request adds the formula for the Pimple DI container -- http://pimple.sensiolabs.org/


alanthing:
Could you [add a test](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#add-a-test-to-the-formula) for each formula so it passes the build?

noodlehaus:
ok, working on it.

noodlehaus:
hi, have added test, but in the build, it gives this.

```
==> Installing php56-pimple from Homebrew/homebrew-php
Error: You must `brew link php56' before php56-pimple can be installed
```

this seems to be an issue with the CI server, or please let me know if there's anything else i need to do, otherwise. thanks!

alanthing:
Thanks. I think the failure is something specific to new formulas within homebrew-php. I'll investigate, but this builds cleanly.

PR #1516, opened 2015-01-23
Bump php versions to 5.4.37, 5.5.21 and 5.6.5

dguyon:
Following http://php.net/archive/2015.php#id2015-01-22-3


alanthing:
Thanks!

PR #1522, opened 2015-01-27
drupal-code-sniffer 8.x-2.0

arnested:


alanthing:
I see this is a 8.x release; is still backwards compatible with Drupal 7 and/or 6? If it's not, we may want to offer 7.x by default and provide 8.x as a devel/head option since Drupal 8 is not released yet. If that's not the case and it is backwards compatible, we can get this pushed through.

arnested:
This is backwards compatible.

arnested:
I also added the DrupalPractice standard which is now part of Drupal Coder Sniffer.

This time the test build went fine.

alanthing:
Great, thanks!

PR #1526, opened 2015-02-02
abstract-php-extension: use opt_prefix instead of prefix for module location

alanthing:
Scenario: I installed php54-opcache and it created a new file `$(brew --prefix)/etc/php/5.4/conf.d/ext-opcache.ini` containing "`zend_extension="/usr/local/Cellar/php54-opcache/7.0.3/opcache.so"`". Great, everything works. Later, I bump the version to 7.0.4 in the formula file and `brew upgrade php54-opcache` and the new file is at `$(brew --prefix)/Cellar/php54-opcache/7.0.4/opcache.so`, but my `ext-opcache.ini` file still contains the hard reference to the *7.0.3/opcache.so* file. 

Since a symlink is created in `$(brew --prefix)/opt/` for each PHP extension, I think it might make more sense to use this `opt_prefix` path instead of the hard path. That way, you can install your extensions and get PHP going how you like, and upgrades keep things sailing with new versions as you'd expect. 

Any objections, @Homebrew/php ?


alanthing:
I tested going from 7.0.3 to 7.0.4 without this fix and the ext-opcache.ini file was updated to the 7.0.4 path, keeping up with the new version. But, when I did it last week, it didn't happen; I had edited the file to make some custom changes that time and perhaps something with regex replacement rules didn't jive (this time the file was not touched at all)? Either way, using the `opt_prefix` would be more foolproof, shifting the task of keeping the [zend_]extension path updated to `brew`'s symlinking in `$(brew --prefix)/opt` instead of us needing to keep it updated. In light of this, I'm still interested in hearing if there objections to this change. Thanks.

PR #1534, opened 2015-02-11
Parsekit patch won't apply

SteelPangolin:
The existing patch is correct, but can't be applied due to whitespace changes. I've updated the patch so that it applies cleanly.


SteelPangolin:
Reverting the patch to just 5.4, since 5.5 and 5.6 are failing builds.

PR #1543, opened 2015-02-20
phpctags v0.5.1

meunierd:
I'm not sure if this is the best way to package it, but I'd be happy to fix it if you have any suggestions.


alanthing:
Looks good, thanks!

Issue #1223, opened 2014-06-29
Missing bottles


mcuelenaere:
homebrew-php doesn't have bottles for any of its formulae.

This for example slows down Travis builds as it has to compile PHP from source.

zimme:
With homebrew/homebrew#30540 merged, is this a possibility yet?

mikemcquaid:
It is but uploading bottles of new taps is temporarily broken. It will be completely ready in a week or so I suspect.

alanthing:
For what it's worth, it doesn't look like we can build a bottle for php5x because the default options rely on the system OpenSSL:

```bash
$ brew test-bot --local --tap=homebrew/homebrew-php php56
... snip ...
  /usr/local/Cellar/php56/5.6.6/include/php/Zend/zend_vm_def.h
  /usr/local/Cellar/php56/5.6.6/include/php/Zend/zend_vm_execute.h
  /usr/local/Cellar/php56/5.6.6/include/php/Zend/zend_vm_opcodes.h
Error: 2 problems in 1 formulae
 * object files were linked against system openssl
These object files were linked against the deprecated system OpenSSL.
Adding `depends_on "openssl"` to the formula may help.
  /usr/local/Cellar/php56/5.6.6/bin/php
  /usr/local/Cellar/php56/5.6.6/bin/php-cgi
  /usr/local/Cellar/php56/5.6.6/libexec/apache2/libphp5.so
```

It seems like the resolution would be to require the Homebrew OpenSSL, but I'm not sure if that would be desired. I'd be anxious to hear others' thoughts about this.

alanthing:
There's also this section about shadow header files, which I don't understand:

```bash
==> brew audit php56							 FAILED
php56:
 * Header files that shadow system header files were installed to "/usr/local/Cellar/php56/5.6.6/include"
The offending files are:
  /usr/local/Cellar/php56/5.6.6/include/php/ext/date/lib/timelib.h
  /usr/local/Cellar/php56/5.6.6/include/php/ext/date/lib/timelib_config.h
  /usr/local/Cellar/php56/5.6.6/include/php/ext/date/lib/timelib_structs.h
  /usr/local/Cellar/php56/5.6.6/include/php/ext/date/php_date.h
  /usr/local/Cellar/php56/5.6.6/include/php/ext/dom/xml_common.h
  /usr/local/Cellar/php56/5.6.6/include/php/ext/ereg/php_ereg.h
[...]

PR #1544, opened 2015-02-22
Changes for bottling default formulas.

alanthing:
If we want to have bottles be the default distribution method:

* Default to build both mod_php and PHP-FPM
* Use Homebrew OpenSSL by default and remove option to use built-in OpenSSL
  (http://pecl.php.net/~jani/patches/multi-sapi.patch)
* Add execute bit to PHP-FPM control script
* Make '#{bin}/phar' symlink relative instead of absolute
* Change a "system 'rm ...'" call to FileUtils.rm for a `brew audit` warning
* Remove "--enable-fastcgi"; it was changed to "--enable-cgi" in PHP 5.3
  (http://php.net/manual/en/configure.about.php)
* Remove "def build_apache?" and create "def build_fpm?" to better suit
  building multiple SAPIs at once but not FPM and CGI since they collide
* Back-port patch for building multiple SAPIs at once to PHP 5.3

We can cover both major installation methods with a distributed bottle. This works because the Apache module and PHP-FPM will not clash when not distributed by a bottle because the `apxs` installation will not fire (that's part of `php`'s `make` procedure). And, if one prefers mod_php to PHP-FPM, the `php-fpm` binary and associated files can simply be ignored, Or, one can always add their own build options to build their own source version and bypass the bottle. You can technically use `--without-apache --without-fpm` and only get the PHP binary, or you can use `--with-cgi` if you don't like PHP-FPM. Using `--with-cgi` implies `--without-fpm` since you can't build two [F]CGI binaries.


alanthing:
One early issue that needs remedying: looks like at least php53 won't build both mod_php and PHP-FPM at once. I did my preliminary testing with only php56. I'll revisit this and any other issues later.

DomT4:
> Default to Homebrew OpenSSL with option to use built-in OpenSSL

Honestly, I'd seriously consider scrapping even the option to use the system OpenSSL. It's pretty horribly unsupported by Apple, and at the end of 2015 the OpenSSL 0.9.8x branch Apple has clung to is going to stop receiving security updates from OpenSSL themselves, several of which Apple has *already* missed. I think you'd make life easier (and more secure) if you mandated the use of brewed OpenSSL.

alanthing:
That's great feedback @DomT4. I'll revisit abstract-php.rb and the php5x.rb formulas to do that and update the PR branch. Effect on users should be minimal with Homebrew OpenSSL being distributed via bottle.

alanthing:
I'm working on a way to get PHP 5.3 to build both mod_php and PHP-FPM at the same time.

But, because `brew audit` will always fail the php5x formulas due to the header files, we'll never see a "pass" build for the main formulas from the Jenkins brew bot. Bummer.

josegonzalez:
@alanthing Care to expand on that comment about header files?

alanthing:
@mikemcquaid touched on it at https://github.com/Homebrew/homebrew-php/issues/1223#issuecomment-75378340. Basically, we need to include header files so people can use `pecl` and `phpize && ./configure && make` etc to build PHP extensions. The `brew audit` code marks including headers as an issue because they contain 'shadow' references to non-brew files, like the Xcode CLT libraries and headers. We can still build bottles successfully, but GitHub will always show that the Jenkins tests Failed due to this audit failure. Unless there's a way to add a flag somehow to ignore that component of `brew audit`...

Unrelated, I will squash this PR down to one commit when it's completed.

mikemcquaid:
> Unless there's a way to add a flag somehow to ignore that component of brew audit...

There isn't, sorry :(

alanthing:
I think it'll be okay. Aside from bumping versions, the php5x formulas themselves change infrequently. We can always run the brew test-bot commands locally, and/or parse through the specific failures and know we're good if it's only the shadow header bit that failed.

DomT4:
> There isn't, sorry :(

Could we ` return ` or similar on the audit check for shadowed headers for the ` php ` formulae? I think we do for Postgres/Libtool already. Unless the ` php ` warnings are useful for debugging, etc.

alanthing:
I've updated the PR to include bottles. Would folks be willing to test?

mikemcquaid:
@alanthing That's going to behave weirdly because then the bot will try and rebottle the bottles in this PR. I recommend pointing people towards `brew pull` instead of adding bottle blocks to PRs.

alanthing:
I added it because it didn't look like the test bot had built bottles:

```bash
$  brew pull --bottle https://github.com/Homebrew/homebrew-php/pull/1544
######################################################################## 100.0%
==> Applying patch
Applying: Changes for bottling default php5x formulas:
==> Patch closes issue #1544
Error: Failure while executing: /usr/bin/curl -f#LA Homebrew\ 0.9.5\ (Ruby\ 2.0.0-481;\ Mac\ OS\ X\ 10.10.2) --silent --fail -o /dev/null -I https://github.com/BrewTestBot/homebrew-php/compare/homebrew:master...pr-1544
```

mikemcquaid:
You'll need to check the downstream build view to work out why they didn't upload: http://bot.brew.sh/job/Homebrew%20PHP%20Pull%20Requests/124/downstreambuildview/

alanthing:
@mikemcquaid Looks like it's something outside of my control:

```
Warning: Could not create link for homebrew/games/omega, as it
conflicts with Homebrew/homebrew/omega.
```
http://bot.brew.sh/job/Homebrew%20Bottles/16400/console is for a build where I removed the bottle URLs and pushed. The word 'omega' is not anywhere in the homebrew-php repo.

mikemcquaid:
You can ignore that warning. That build was from before the tap bottle issues were fixed.

alanthing:
What should I be seeing then for build artifacts for this job? #1545 had bottles build but I don't see where the problem lies on builds that do have completed bottles and ones that don't.

alanthing:
I put the bottle references back in since I had asked people to test (I built everything manually in VMs and added to BinTray by hand). Let me know if you'd like me to remove them and re-push.

mikemcquaid:
Seems to be behaving now: http://bot.brew.sh/job/Homebrew%20PHP%20Pull%20Requests/125 Probably want to remove the bottle blocks here, yeh.

alanthing:
Thanks for your help @mikemcquaid. Last question: for BinTray, I have the two environmental variables on my local. What does the test bot do with BinTray at the conclusion of a job, if anything? And, since I manually created the php5x packages and versions earlier, should I completely delete them before I run `brew pull --bottle`?

mikemcquaid:
> Thanks for your help @mikemcquaid. Last question: for BinTray, I have the two environmental variables on my local. What does the test bot do with BinTray at the conclusion of a job, if anything? And, since I manually created the php5x packages and versions earlier, should I completely delete them before I run brew pull --bottle?

So, the test bot uploads everything to SourceForge but if it doesn't have an existing bottle it is not yet published. Bintray deletes anything unpublished in a repository after 24 hours. When you `brew pull --bottle` your API key is used to tell Bintray to publish any unpublished files for that formula. I'd recommend generally not manually uploading packages to Bintray as I'm not sure what the exact effects will be on the uploading flow.

PR #1552, opened 2015-02-25
Added php56-ioncubeloader formula

Zifius:
Ref #1550


Zifius:
@alanthing first time adding a formula, should I ignore failed build complaining about code style checks? Bottles seem to be created

josegonzalez:
@Zifius I would fix code style violations, since they are so small, should be easy to do them :)

Zifius:
@josegonzalez thanks for the reply and no problem with Vim :)

Zifius:
@josegonzalez code style checks passed but now I'm asked about tests (probably because of --strict flag), what gives?

josegonzalez:
@alanthing thoughts?

Zifius:
Ok found tests in other formulas but got this error now: "PHP Fatal error:  [ionCube Loader] The Loader must appear as the first entry in the php.ini file in Unknown on line 0"
Though locally it installs fine and php -m works

alanthing:
It's failing on 5.5 and 5.6 because we're using `zend_extension` in php.ini for opcache, ala: https://github.com/Homebrew/homebrew-php/blob/199bf6d/Abstract/abstract-php.rb#L356

When I made that change, I made the argument that, because PHP's default './configure && make && make install' adds *and enables* opcache, that we should too. Which, you might be wondering,: if it's the default, then why do we need to enable it post-install? We're using `php.ini-development` as our source `php.ini` file, which has opcache explicitly disabled.

If we moved the opcache config to `conf.d/ext-opcache.ini`, it would solve this problem because `conf.d/ext-ioncubeloader.ini` would come first alphabetically, and therefore resolve this issue, but it doesn't solve the problem of another zend_extension, like `conf.d/ext-afakething.ini`, causing other problems down the road for users. Brew test bot problems would go away indefinitely though.

We could set `--disable-opcache` in the `php55` and `php56` formulas, create `php55-opcache` and `php56-opcache` formulas, and move the configuration to `conf.d/ext-opcache.ini`, which would pass this ionCubeLoader test and any other extensions that have the same problem. If users had ionCubeLoader and any pre-"i" zend_extensions, they would need to manually rename `conf.d/ext-ioncubeloader.ini` to something like `conf.d/a-ioncubeloader.ini`, but we shouldn't make that decision for people by default. Maybe mention it in the caveats/info?

While I'm not a huge fan of replicating built-in functionality as an extension, since opcache ships with PHP 5.5 and 5.6, it makes using opcache **opt-in** without requiring a rebuild of the php55 and php56 bottles (because adding `--disable-opcache` forces a source build). It also allows for newer versions of opcache to be used: 7.0.3 ships with PHP 5.5, but 7.0.4 is available for 5.4 and 5.3.

**TL;DR**: this is resolved by disabling opcache as a default option for `php55` and `php56` and moving it to to-be-created `php55-opcache` and `php56-opcache` formulas.

Thoughts?

Zifius:
Thanks Alan, at least it explains why things working locally for me - I still have the /usr/local/etc/php/5.5/conf.d/ext-opcache.ini and the extension is enabled through this file

Otherwise the changes you proposed seem to be the way to go, except that we should consider people not paying much to some major changes (like it was in my case - I saw the opcache related commit but didn't change anything in my local configuration)

Zifius:
@alanthing should we rerun the build to see if the tests are green since you committed the changes in #1553 to master?

alanthing:
Sure, I can't manually fire tests but any changes to your branch will trigger it. You could simply `git commit --amend`, edit your commit message, and `git push --force` and it should retry.

alanthing:
Because the Brew bot does not wipe the build environment when multiple formulas are tested, I think it might be best to break these up. Bottles are the way to go forward, and we should get in the habit of one pull request for each formula. It's a little more work, but it drastically increases the likelihood of passing tests and getting the bot to build bottles for us.

Would you be willing to 1) create php5{5,4,3}-ioncube branches and submit new PRs, and 2) make this PR only contain changes to php56-ioncube?

Zifius:
@alanthing sure thing, done now. PS. it looks like the build system is kind of hung up

alanthing:
@BrewTestBot test this please

alanthing:
```bash
==> Downloading http://downloads3.ioncube.com/loader_downloads/ioncube_loaders_dar_x86-64.tar.gz
Already downloaded: /Library/Caches/Homebrew/php56-ioncubeloader-4.7.4.tar.gz
==> Verifying php56-ioncubeloader-4.7.4.tar.gz checksum
Error: SHA1 mismatch
Expected: f4bf288101d2aa3ec4ebfcfdc6044eeaf0ae9690
Actual: ca4f6c68f05e4b7255af257d32198be5a61b2604
Archive: /Library/Caches/Homebrew/php56-ioncubeloader-4.7.4.tar.gz
```

The URL http://downloads3.ioncube.com/loader_downloads/ioncube_loaders_dar_x86-64.tar.gz looks like it's a "latest download here" URL and will stay at the same location when they release new versions. That doesn't jive well with Homebrew since we require SHA checksums of all downloaded files. Does IonCube provide version-specific download URLs? If so, we should use that so we don't end up with checksum failures, and we can manually bump versions as IonCube releases them.

Zifius:
@alanthing I did not found version specific on the site (the listing for the directory is forbidden), but will fill a support query and ask. Meanwhile subscribed to loader updates so in case they release a new version I can send a PR in a reasonable timeframe. Closing this PR and will resubmit a new one as I messed up parent commit of this branch in my local copy

PR #1553, opened 2015-02-25
Disable opcache in php5[56].rb in favor of php5[56]-opcache.rb

alanthing:
Having the opcache `zend_extension` in php.ini can be problematic in certain situations. Forcing opcache by default, while a default option for upsteam PHP's `./configure`, may not be desirable in certain development configurations. Also, this makes it possible to use newer versions of ZendOpcache than bundled with PHP 5.5 or 5.6.

This removes `--enable-opcache` from `php55` and `php56` in favor of using `php55-opcache` and `php56-opcache`.

See further discussion in #1552.


alanthing:
Unfortunately, this test will continue to fail (aside from shadow header files) because I have multiple `php5x.rb` changes alongside multiple `php5x-opcache.rb` changes. Any `php5x-opcache.rb` test install other than the first will fail because it will need `brew unlink php5x ; brew link php5y` to be run first.

Lesson-learned for later: don't combine PHP and extension formulas for multiple PHP versions in the same PR. I'll use fresh VMs for building new bottles for the opcache formulas if I need to.

PR #1557, opened 2015-02-26
pecl-jsmin updates for 5.3, 5.4, 5.5 and 5.6

jwhitcraft:
- Update to the latest version of the pecl jsmin extension which is 1.1.0


jwhitcraft:
going to close this to move each one into it's own PR so it will go green.

alanthing:
Glad you noticed to do that! Will merge here in a minute...

PR #1568, opened 2015-02-28
Add php56-ioncubeloader formula

Zifius:


Zifius:
@alanthing the IonCube support responded that they don't keep old version on purpose but they offered a checksum feed implementation, though I'm not sure that this is supported in homebrew (I'm primarily PHP developer). If you can shed some insight on how to implement it that would be appreciated

alanthing:
Try removing the `sha1` portions. If it builds locally for you, commit and push and we'll see what the brew bot thinks. It might fail in the audit section, but that'd be okay if it built successfully and we knew it was acceptable for this formula.

Zifius:
@alanthing thanks, php55 built green other versions just resubmitted

alanthing:
Thanks for your work on this!

Zifius:
@alanthing cheers Alan, thanks for your guidance

PR #1580, opened 2015-03-08
Add uopz 2.0.7

jk:
This PR adds the User Operations for Zend (uopz) extension to all supported PHP versions:
* 5.4
* 5.5
* 5.6

There is no 5.3 support since uopz is incompatible to 5.3 as stated https://github.com/krakjoe/uopz and http://php.net/manual/en/uopz.requirements.php.


alanthing:
Could you review the test result and address the output of the audit test and make those changes?

Also, the build tests are significantly more likely to pass if you create a pull request for each formula instead of bundling them into one. Could you split these into 3 PRs? The [hub](https://github.com/github/hub) tool can speed up the process.

jk:
@alanthing I'll do the requested changes.

One thing: the uopz extension should be loaded before opcache if present. How can I make sure this is the case? Is there a directive to ensure that?

alanthing:
For the purposes of this formula and building tests, this won't be a problem because the current versions of `php55` and `php56` no longer include *opcache.so* in the build result as it was causing some trouble. 

Unfortunately, there's not a built-in way to deal with opcache and the order of loading zend_extensions. Therefore, you should add a note in the formula to display after installation about the order or extensions as relevant specifically to uopz. For example, since this needs to load before opcache, mention that `/usr/local/etc/php/5.6/conf/ext-opcache.ini` should be renamed to be after `ext-uopz.ini` alphabetically, or the other way around. 

I'll keep this mind with other `zend_extensions` from now on, and if it seems like opcache *always* needs to be loaded last, we can look into having it's configuration file be `opcache.ini` instead of `ext-opcache.ini` or similar.

PR #1588, opened 2015-03-10
php56-xdebug: revert to 2.2.7 until 2.3.2 is released

alanthing:
See http://bugs.xdebug.org/view.php?id=1122


alanthing:
@BrewTestBot test this please

Issue #1598, opened 2015-03-10
Installing php55-redis with igbinary


myilmaz:
Trying to install php55-redis --with-igbinary

```
==> Installing php55-redis from homebrew/homebrew-php
==> Downloading https://github.com/phpredis/phpredis/archive/2.2.7.tar.gz
Already downloaded: /Library/Caches/Homebrew/php55-redis-2.2.7.tar.gz
==> /usr/local/opt/php55/bin/phpize
Error: No such file or directory - ext/igbinary
```

PR #1599, opened 2015-03-10
php55-redis: installing --with-igbinary

myilmaz:
Closes: #1598.


alanthing:
Is there any reason why we wouldn't want `php55-igbinary` included by default? It's a hard requirement of `php-pecl-redis` on CentOS, and I don't think it's a problem to make the two linked here in Homebrew too. Your thoughts? That would solve your issue here and enable the use of bottles when needed redis with igbinary.

alanthing:
Thanks for this! I cherry-picked your commit into my other PR so you'll still be listed in the commits :trophy:

PR #1625, opened 2015-03-22
Update xdebug to 2.3.2

j0k3r:
http://xdebug.org/updates.php#x_2_3_2


localheinz:
:+1:

alanthing:
The 2.3.x branch caused a ton of problems the last time we had it in. Let me do some testing before we accept this PR.

j0k3r:
@alanthing yep, I also got a lot of problem with some segfault on coverage. I'm using the 2.3.2 since few days now without any problems.

alanthing:
Ah, so it was 2.3 and 2.3.1 that had issues... I'll revisit this soon. Thanks for the feedback @j0k3r.

PR #1622, opened 2015-03-20
Bump php versions to 5.4.39, 5.5.23 and 5.6.7

dguyon:
Following http://php.net/archive/2015.php#id2015-03-20-2

PR #1626, opened 2015-03-23
Build new bottles for php54, php55, and php56

alanthing:
Thank you, @dguyon, for bumping the PHP versions in #1622. By also editing the php5x.rb formulas, the build bot will create and upload new binary bottles. That is the sole purpose of this PR.


alanthing:
@BrewTestBot test this please

PR #1630, opened 2015-03-24
remove disable-debug configure option for xdebug

vimac:
xdebug configure flag "--disable-debug" will cause incompatible with php which enable debug flag


alanthing:
Can you fix so it passes the build tests?

vimac:
Which is better, remove so from bintray in *xdebug.rb or recompile so?
BTW, I can't access the merged build failed detail page with 404 error (http://bot.brew.sh/job/Homebrew%20PHP%20Pull%20Requests/224/)

I'm a newbee for contributing homebrew project :)

alanthing:
Closing as I have removed the bottles. Due to the variances of xdebug and PHP and whether or not certain debug options should be used when building binaries, it seems ideal to alway shave xdebug build from source.

PR #1662, opened 2015-04-16
Adds php-sodium formulae

cmckni3:


alanthing:
The tests are more likely to pass if each new formula is submitted as a separate pull request. Because you've created a formula that depends on another, I'd recommend creating 4 new pull requests for each `php5x-libsodium.rb` formula, we'll merge if it passes tests, and then we'll repeat for the `php5x-sodium.rb` formulas afterwards. 

Also please ensure each pull request is a single commit. That also increases the likelihood of passing the brew test bot.

The [hub](https://github.com/github/hub) tool can really speed up submitting PRs.

cmckni3:
@alanthing I can do that for debugging purposes. It's strange that an encoding error occurs.

    Error: invalid byte sequence in US-ASCII

cmckni3:
Closing because I started making individual pull requests.

PR #1687, opened 2015-04-28
wp-cli v0.19.0

oknoway:


cmckni3:
:+1:

danielbachhuber:
@oknoway @cmckni3 Is it worthwhile to add anything to our [release checklist](https://github.com/wp-cli/wp-cli/wiki/Release-Checklist) to ensure Homebrew uses the latest version?

cmckni3:
@danielbachhuber I think you could add a new section to the release checklist to update the `homebrew-php` formula for `wp-cli`. That's completely up to the maintainers of `wp-cli` though. Someone will eventually send a PR to `homebrew-php` to update `wp-cli`.

I thought about some sort of way to make `homebrew` update to the latest without having to edit the formulae. I think it goes against the way `homebrew` itself works though since it verifies the sha1 or sha256 of the downloaded file to make sure it wasn't tampered with while downloading.

The latest version from `master` can still be installed using `--HEAD` but that is only helpful when `master` is used only for production releases.

Overall, I don't think the workflow of creating brew formulae should be changed.

danielbachhuber:
> I think you could add a new section to the release checklist to update the homebrew-php formula for wp-cli.

Sounds good — I've done so.

> That's completely up to the maintainers of wp-cli though.

That's me :)

> Someone will eventually send a PR to homebrew-php to update wp-cli.
> [...]
> Overall, I don't think the workflow of creating brew formulae should be changed.

Seems reasonable. A PR is fine — I was just ignorant of the process to begin with.

Thanks!

cmckni3:
Awesome, @danielbachhuber!

Haha, I saw you were a maintainer of `wp-cli`. I just wasn't sure of the decision making process for the project.

It is super easy to update a formula. Fixing formulae, not so much. I have had to fix a few formulae recently.

PR #1699, opened 2015-04-30
Adds php56-libsodium

cmckni3:


alanthing:
These build errors are pretty weird, they fail around 'brew cleanup' and stuff like that- tasks related to the actual formula are fine. Anyway, go ahead and submit the php5x-sodium formulas now the same way and I'll get them through.

PR #1702, opened 2015-04-30
Update php-cs-fixer to 1.7

razmo:
Update to the latest version of php-cs-fixer


Glavin001:
Does the checksum sha1 have to also be changed?

razmo:
Yes, forgot it !

Issue #1701, opened 2015-04-30
** README - update icu4c (and possibly others) if you update PHP


alanthing:
**EDIT:** This is resolved. The `intl` extension has been removed from the main `php5x` formulas and moved into `php5x-intl`. If you are experiencing this problem (replace `php56` with `php55`, `php54`, etc as needed)...

* Uninstall your existing Homebrew PHP version if it is the latest release from [php.net](http://php.net)

   `brew uninstall --force php56`
* Update your formulas: 

   `brew update`
* Fix any tap link errors: 

   `brew tap --repair`
* Purge any previous bottle downloads: 

   `find $(brew --cache) -mindepth 1 -print -delete`
* Install your `php5x` version (remember, both `--with-apache` and `--with-fpm` are included by default and do not conflict, so you don't need to include or exclude either of those options, and you'll get the binary bottle, saving a *lot* of compile time)

   `brew install -v php56`
* **Optional: If you need the `intl` extension**: 

   `brew install php56-intl`

Leave a comment in this issue if you need assistance, or create a new issue.

---

**Original issue:**

The binary bottles for the latest versions of php54, php55, and php56 were built with the newest versions of dependencies. On my system, I had to `brew update && brew upgrade icu4c` to resolve an issue with a dynamic link for `/usr/local/opt/icu4c/lib/libicui18n.55.dylib` when I only upgraded `php56`. Depending on how others' systems are configured, you may need to upgrade additional components as well. While it may take a while, a `brew update && brew upgrade` of all system components might be the easiest way to resolve.

I'll leave this issue open for a few weeks to allow folks to find it before opening new issues. Please add your comments if you come across other related issues, or if it's not related, create a new issue.

RandomDSdevel:
Could [this](https://github.com/Homebrew/homebrew/issues/39221) be related?

josegonzalez:
Can we require a specific version of a formula?

alanthing:
@RandomDSdevel I think so. If you update icu4c you should be alright.

alanthing:
@josegonzalez I've not come across any version requirements like Debs and RPMs that stuck with me, but I could be wrong. Homebrew's approach is "run the latest all the time" which makes situations like this a bummer. Hopefully, there's a little expectation of "bleeding edge" if you're upgrading some packages, and it shouldn't be too bad to just upgrade a few more.

DomT4:
[...]

PR #1719, opened 2015-05-11
abstract-php: remove intl options by default

alanthing:
See [#1701](https://github.com/Homebrew/homebrew-php/issues/1701).


alanthing:
I'll add the `php5x-intl` formulas once this is passed and merged.

PR #1730, opened 2015-05-13
Update and Fix Phpmetrics

sofakartoffel:


alanthing:
Thanks! I cleaned up the results from `brew audit --strict` before pushing.

sofakartoffel:
Thank you.

PR #1734, opened 2015-05-15
php5x: remove old bottle definitions

alanthing:
And the test bot should kick off new bottle builds!


josegonzalez:
Heh :)

alanthing:
It's an ironic cycle- formulas need to be edited to build bottles, and the best way to do it is to... remove the old one!

alanthing:
By the way, once the new bottles are built and published, I'll create three other PRs to have extensions relying on the latest source version to get pushed through too.

alanthing:
New bottles built for major versions and the 8 formulas that are built from major source releases.

For future reference: when major PHP versions are updated via the abstract-php-versions.rb file, separately edit each php5x.rb file and remove the bottle hashes but leave the "bottle do ... end" block, and create a PR. Once the PR is run through the brew test bot, run the following on your local: `brew pull --bottle https://github.com/Homebrew/homebrew-php/pull/xxxx`. If all is well, `git push` and that's it.

PR #1750, opened 2015-05-18
Parsekit for PHP 5.5 and 5.6

SteelPangolin:
I've updated the formulas to use the same patch from #1534 that allowed Parsekit to build against PHP 5.4 on PHP 5.5 and 5.6.

Additionally, there is a patch from @Hubbitus on the PHP bug tracker that allows Parsekit to build against PHP 5.6, so I've included that as well.

I've been able to compile and use Parsekit with both 5.5 and 5.6 on my desktop, but last time I submitted an update to these formulas, there were some CI failures, so some extra work may be required.


SteelPangolin:
CI passed this time. Hooray!

alanthing:
Thanks for this! Would you be able to update the PR branch with similar formulas for PHP 5.3 and 5.4? Judging by the patch comment and the age of the pecl releases, I bet it would work.

SteelPangolin:
@alanthing: I previously updated the PHP 5.4 formula for Parsekit in PR #1534 (actually, you were the one who signed off on it).

There's no formula for Parsekit on 5.3, so I've created one and added it to the PR branch. It doesn't need any patches, probably because 5.3 was current when Parsekit development stopped.

SteelPangolin:
Looks like two separate issues: a `brew audit` failure, which I can fix, and some kind of conflict between PHP 5.3's `pear` binary and those for other PHP versions, which I can't, but might possibly result from the 5.3 version failing.

I'm going to update all the Parsekit formulas to meet `brew audit` standards and see if that helps. A successful run on the 5.3 formula might allow Jenkins to clean up after itself.

SteelPangolin:
Don't merge that even if the tests pass. Pretty sure I made a Ruby mistake in the test blocks.

SteelPangolin:
@alanthing: I can't figure out what's going on here with the most recent test failures, and can't reproduce them locally. Parsekit includes some example code in its source tarball, and if the examples run correctly, Parsekit is working. I tried to include them in the install prefix so that they'd be available to the `test` block.

Is there some reason `Dir.chdir prefix` wouldn't work inside a `test` block on the Brew bot? Should I just drop the attempt to run the Parsekit examples as a test, and trust that if it shows up in `php -m` it'll work?

SteelPangolin:
Looks like my mistake: some formulas were missing the line that installs the test scripts.

SteelPangolin:
@alanthing: OK to merge now.

alanthing:
Excellent, thank you! I squashed it down to 1 commit before pushing.

Issue #1751, opened 2015-05-19
Can we have a 2.x pecl-php-coucbase for php54-couchbase?


polesen:
Would it be possible to upgrade the formula for php54-couchbase to build a newer 2.x php-pecl-couchbase? Currently, it seems to be building 1.2.2.

Or should that be a completely new formula, to avoid breaking someone?

alanthing:
In my opinion, as long as 2.x is stable, the formula should be using the latest stable version. If you want to update the paths, sha256s, etc, and submit a PR, we'll let the Brew Test Bot run through it and see how it goes!

PR #1752, opened 2015-05-20
Upgraded php-couchbase module to 2.0.7 version

polesen:
@alanthing here is a PR for issue #1751 

For now, I have only updated php54-couchbase. Maybe it should be done for 53, 55 and 56 too, but I am not running those versions, so I am unsure about them working?

### DIFF
--- a/Formula/php53-couchbase.rb
+++ b/Formula/php53-couchbase.rb
@@ -3,9 +3,9 @@ require File.expand_path("../../Abstract/abstract-php-extension", __FILE__)
 class Php53Couchbase < AbstractPhp53Extension
   init
   homepage 'http://pecl.php.net/package/couchbase'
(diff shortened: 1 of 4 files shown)


alanthing:
Thanks! It'd be great if you could include the others. Basically, save your php54-couchbase.rb file as php55-couchbase.rb, find and replace "54" -> "55", "5.4" -> "5.5", etc. The Brew Test Bot will make sure you've got everything written properly.

polesen:
@alanthing I've applied changed to 53, 55 and 56 too now.

alanthing:
Thanks! I added a test for each one per `brew audit --strict`. Also, I looked at the documentation for https://github.com/couchbaselabs/php-couchbase/, and it looks like the configuration options changed in the 2.x branch, so I removed some options. I don't use Couchbase, so let us know if you encounter any trouble.

PR #1740, opened 2015-05-15
php5x-http: update to 2.4.3

alanthing:

PR #1757, opened 2015-05-25
Update PECL HTTP extension to 2.4.3

ondrejmirtes:


alanthing:
Duplicate of #1740, which apparently needs some patches and I haven't had a chance to resolve.

alanthing:
FYI- got #1740 worked out finally and pushed today.

PR #1759, opened 2015-05-28
add an option to build against homebrew's libxml

benjaminschulz:
I think it's desirable to support building of PHP against homebrew's libxml2, similar to the existing --with-homebrew-libxslt. Yosemite has libxml2 2.9.0 but the current version is 2.9.2.


alanthing:
Thanks!

PR #1760, opened 2015-05-28
Updated Codeception to v2.0.14

Zifius:


alanthing:
Thanks! I added a desc, removed the first line, and reformatted the test as per `brew audit --strict`.

PR #1762, opened 2015-05-29
Updated PECL HTTP to 2.4.3

Vrtak-CZ:


alanthing:
Duplicated of #1740 which needs to have build problems resolved.

alanthing:
FYI- got #1740 worked out finally and pushed today.

PR #1807, opened 2015-06-19
Updated Xdebug to 2.3.3

Zifius:


Zifius:
@alanthing Hi, what could be wrong here in the build?

alanthing:
@Zifius looks like problems with the bot and switching between multiple versions; it should be alright to push through.

alanthing:
I added "desc" as per `brew audit --strict`... thanks!

PR #1813, opened 2015-06-24
add initial php 7.0 support

ablyler:
This also includes the extensions that are part of PHP proper.  It does not include any third party extensions, as most of them do not support PHP 7.0 yet and fail to compile.


alanthing:
Can you view the results from the test? If so, there are a handful of audit/style stuff if you'd like to clean it up a little bit before posting. Otherwise, this looks great Andy. I know a lot of people have been interested in this.

ablyler:
@alanthing: Yep! Looks like there are a bunch of formula's that won't pass these tests already in the repo. I went ahead and cleaned up all but one warning.  It is in regards to: `Header files that shadow system header files were installed to`, which I'm not sure how we'd clean that up.

alanthing:
Yeah, I'm trying to resolve things as they come in, usually by just doing `brew audit --strict path/to/formula.rb` but I haven't gone back and done a lot of the existing ones.

The header files thing has to be ignored. Most Homebrew formulas shouldn't have compiled headers and such, but we need them for `pecl`, `phpize && ./configure` and other things, so that's the one test we can safely ignore.

So this looks good! Thanks! I'll merge and get the bottles added via `brew pull --bottle https://github.com/Homebrew/homebrew-php/pull/1813 && git push`

PR #1817, opened 2015-06-29
Updated Codeception to v2.0.15

Zifius:


Zifius:
Will resubmit with 2.1 released today

PR #1818, opened 2015-06-30
Updated Codeception to v2.1.0, added caveats

Zifius:


Zifius:
>Parse error: parse error in phar:///usr/local/Cellar/codeception/2.1.0/libexec/codecept-2.1.0.phar/vendor/guzzlehttp/psr7/src/functions.php on line 77

The line 77 reads:
```php
function stream_for($resource = '', array $options = [])
```

Probably Mountain Lion build is tested with PHP which doesn't support short array syntax. Is it issue for us @alanthing?

Zifius:
@alanthing @josegonzalez Hey guys, according to the author of the Codeception since version 2.0 PHP 5.3 is not supported. Do you know a way to disable testing this Formula on 5.3?

ablyler:
Merged, thanks! There currently isn't a way to disable the tests on a specific version of PHP.

PR #1819, opened 2015-07-02
Add Adminer from http://www.adminer.org/.

wedi:
Adminer (formerly phpMinAdmin) is a full-featured database management tool written in PHP. It's a lightweight and easy alternative to phpMyAdmin.

It's just one file but having it in homebrew makes it easier to keep track of updates.

The formula is based on phpmyadmin.rb.


ablyler:
Merged, thanks for the PR.

PR #1821, opened 2015-07-02
Fix audit problems in phpmyadmin.rb

wedi:
Fix all audit problems and style complaints reported by `brew audit --strict phpmyadmin`


ablyler:
Merged via 28379c5.

PR #1822, opened 2015-07-03
New formulaes: php70-phan and php70-ast

arnested:
This a formula for phan - Rasmus Lerdofs new [static analyzer for PHP 7](https://github.com/rlerdorf/phan).

phan depends on [php-ast](https://github.com/nikic/php-ast) which is also not in homebrew-php so I added a foumula for php-ast as well.

Unfortunately neither phan nor php-ast has tagged / stable releases so both these formulas are head only.

I'm unsure whether it is a good idea to add head only formulas to homebrew-php but at least this is a good starting point until they one do stable releases.


arnested:
I'm just gonna close the pull request. I don't think having head-only formulaes in homebrew/php is a good idea.

If someone want to try it out I created a repos with just these formulas: https://github.com/arnested/homebrew-phan

PR #1825, opened 2015-07-05
add fast_assert formula for php56

matthew-james:
This PR adds a formula for the [fast_assert extension](https://github.com/box/fast_assert) by Box.

There are 2 caveats:

1.) Box doesn't list compatible versions for this extension.  I have only used it on php56 so that's what I added the formula for.

2.) Box doesn't have any tagged versions, so this is a HEAD only formula.


matthew-james:
Looks like it failed because it's HEAD only.

ablyler:
@matthew-james: can you open an issue with box to get them to tag a version?

matthew-james:
Opened, waiting for a response.

alanthing:
@matthew-james: while we're waiting on tags, I have another question: If this extension particular to PHP 5.6? If not, could you also create formulas for PHP 5.3, 5.4, and 5.5? For the most part, you should be able to just duplicate the file you created and find+replace "56" and "5.6" as appropriate. Thanks!

matthew-james:
I am waiting for an official response on the compatible versions, I will update this PR once I know for sure.

PR #1843, opened 2015-07-11
php5*-oauth: add bottles

ablyler:


alanthing:
For some reason, the test bot can sometimes fail when a single PR needs to span multiple PHP versions for testing. It doesn't seem to be able to always properly uninstall one version before switching between to the next one. In the past, I've recommended that folks do one PR per formula.

PR #1845, opened 2015-07-13
Arcanist Formula

michaeljs1990:
This allows you to use the arc command to work with phabricator.
Hopefully will help cut down on having to manually do this for a lot of
people. It does not have composer support unless you hack it together
using the vcs tags.


michaeljs1990:
Installing arc is a pain point for much of my team. I created this package and i think it's a very good fit into the homebrew-php one. I would actively maintain this as the phabricator team always has a working master branch and it needs to be updated after about ~4 months at the latest likely monthly. Let me know whatever I need to get this in quick :+1: and I'll have it done.

alanthing:
Thanks! If the upstream project changes, please submit new PRs to keep it up to date and we'll be happy to merge.

ablyler:
@michaeljs1990 thanks for the pull request! can you change it to be a HEAD only formula for now? https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#head and can you also ask upstream if they could start using versions? If not, we could always revert back to using the current date or last commit as the version.

The only problem I see with your formula is that users will not be able to upgrade it, because the version will not change.

alanthing:
(Good catch @ablyler)

michaeljs1990:
@alanthing @ablyler thank you for the quick feedback. I believe my commit addressed both of the concerns but I believe to upgrade you will need to do a reinstall still.

michaeljs1990:
@alanthing well that was un ugly rebase but I think the end result is clean :) and the build should pass now...

michaeljs1990:
looks like the build failed but because of the php56 package not mine.

michaeljs1990:
@ablyler awesome! thank you so much for getting this merged in quick.

PR #1846, opened 2015-07-13
Change apache load module instructions as outdated with PHP 7.0

philcook:
Noticed in the output in the console that when installing PHP 7 it has instructions for adding the module in apache for PHP 5 versions.

Created a simple if statement to handle the difference.


ablyler:
Merged, thanks for the PR!

PR #1847, opened 2015-07-13
Add formula for gRPC PHP extension

stanley-cheung:
We are a group from Google working on the next generation open source RPC library and framework with HTTP/2, called gRPC. You can check out our project [here](https://github.com/grpc/grpc) on Github. We have built a gRPC PHP extension via [PECL](https://pecl.php.net/package/grpc) and would like to submit our extension as a homebrew-php formula as well. Thank you.


alanthing:
Hi @stanley-cheung, thank you for this.

* Does gRPC for work PHP 5.4 and/or PHP 5.3 as well? If so, please create additional formulas for those versions if applicable. Even if you feel like adoption would be low, we like to have all available formulas unless in situations where it is impossible to build. For example, the ZendOpcache extension works all the way back to 5.3 as a PECL module; so while most people might not use it, we still provide it because it technically works.
* Please run `brew audit --strict php56-grpc.rb` and follow the recommendations and make corrections in order to pass all of the Brew Test Bot tests; repeat for other PHP major versions as well. For example, it will ask you to provide a `desc` short description.

Thanks!

stanley-cheung:
Hi @alanthing, thank you for the feedback!
 * gRPC PHP is currently in Alpha. So far we have only provided support for PHP 5.5 and above. We are working on making sure we can support 5.4 or even 5.3. 
 * Having said that, our formula may rapidly change over the course of the next few months before we reach GA, with the expected minor version bumps, alongside with our core [gRPC formula](https://github.com/grpc/homebrew-grpc/blob/master/Formula/grpc.rb). Is this a good time to submit this PR knowing we may update the versions along the way?
 * Thanks for the pointer to `brew audit --strict`. I have run this against our 2 formulas and updated them accordingly. Now they passed
 ```
 brew audit --strict php*-grpc.rb
 ==> brew style php55-grpc php56-grpc

 2 files inspected, no offenses detected
 ```

alanthing:
* If you're only supporting 5.5 and 5.6, then that's totally cool. I wanted to make sure we had full coverage *where applicable*.
* Thank you for the updates. Looks like the tests failed because they just contain `false`. Since you're building an extension, try a test like this: [`shell_output("php -m").include?("grpc")`](https://github.com/Homebrew/homebrew-php/blob/master/Formula/php56-yaml.rb#L35)

stanley-cheung:
Thanks for the pointer to the test. I updated the formulas but the Jenkins tests are still failing. I am a little unsure how the tests were set up. Can you see if they are OK? Thanks!

ablyler:
Looks like tests are failing due to the issue I fixed in https://github.com/Homebrew/homebrew/pull/41602

PR #1936, opened 2015-07-24
Remove old bottle definitions for upstream-bundled extensions

alanthing:
Piggy-packing off of the update of PHP 7.0.0 from beta1 to beta2; update the PHP-supplied extension bottles


alanthing:
Thanks @ablyler; the bot was having issues when I was trying to get the bottles built.

PR #2007, opened 2015-08-21
php70: update 7.0.0-rc.1.

JulienBreux:
Missing extensions bottle. How to do ?


ablyler:
@JulienBreux you need to remove the bottle sections from those formulas as well.  I went ahead and did that and merged it in.  thanks!

PR #2031, opened 2015-08-30
Patch 1

Dmitry-Kucher:


Zifius:
Hi, can you split your commit for linuxbrew support?

ablyler:
Merged the linuxbrew support only.

PR #2036, opened 2015-09-06
Bump PHP 5.4 to 5.4.45

Zifius:


ablyler:
Thanks, I've added a commit to regenerate the bottles in PR #2043.

PR #2040, opened 2015-09-08
Fix build with tidy-html5 v5.0.0

kabel:
tidy-html5 starting at v5.0.0 has renamed include files used by the
original lib.


ablyler:
Updated your fix in PR #2049.  Thanks!

PR #2098, opened 2015-09-18
php53-imagick: covert from sha1 to sha256

ablyler:


mikemcquaid:
@ablyler Just an FYI I'm manually killing all these sha1 to sha256 jobs. Ideally they'd just be pushed to `master` in future as they are trivial to verify locally without causing issues for our CI infrastructure.

ablyler:
@mikemcquaid I was also using this to generate bottles for these formulas, I was kicking off 10 formulas at a time and waiting for them to finish before starting the next batch.  Maybe there is a better way to generate bottles for all of our formulas?

mikemcquaid:
@ablyler That's OK to do but it would be good to try and do them 1-2 at a time, thanks.

ablyler:
@mikemcquaid roger, I'll adjust my script. is there a spot you commit homebrew related tools? I've been creating scripts to do a lot of this stuff, and it might be nice to have a toolbox repo for all core and tap devs to use.

ablyler:
@mikemcquaid also, I created el capitan bottles for the core php formulas (php53 - php70), which should decrease the extension build times a ton.

mikemcquaid:
https://github.com/homebrew/homebrew-dev-tools

ablyler:
@mikemcquaid thanks!

PR #2122, opened 2015-09-19
initial work towards splitting the php formulas

ablyler:


ablyler:
@homebrew/php I think this is ready to be merged, anyone want to pull this branch and see what you think?

javian:
I've not studied the code too closely yet but what purpose does php56.rb fill if the functions are split over different formulas ? Maybe it would be good to summarize the changes explaining the intent and purpose and how it would change if that hasn't already been done.

javian:
I'll close this now since the work has begun to merge php into core which would supersede this work.

PR #2193, opened 2015-09-21
php56-xcache: covert from sha1 to sha256

ablyler:


ablyler:
@BrewTestBot test this please

PR #2281, opened 2015-09-25
Update php55-pdo-pgsql.rb

gms8994:
Installing pdo-pgsql should not require a postgresql server installation.


ablyler:
@gms8994 actually it looks like it is required: http://php.net/manual/en/ref.pdo-pgsql.php also the builds failed because they couldn't find the postgresql libraries.

PR #2322, opened 2015-09-27
Travis CI: fix config based on comments from @xu-cheng

ablyler:


ablyler:
@xu-cheng can you review these changes?

xu-cheng:
:+1:

PR #2430, opened 2015-10-02
Bump PHP 5.6 to 5.6.14

Zifius:


ablyler:
@Zifius can you remove the bottle hashes for all formulas that start with php56 and contain `version PHP_VERSION`? Also, remove any `revision` lines from these formulas while you are at it.

Formula/php56.rb
```diff
@@ -12,9 +12,6 @@ class Php56 < AbstractPhp
   head PHP_GITHUB_URL, :branch => PHP_BRANCH
```
ablyler: remove `revision 3` from above

PR #2431, opened 2015-10-02
Codeception 2.1.3

Zifius:


Zifius:
@ablyler for some reason I cannot update with php55, says that formula is dependent on php56:

`==> Installing homebrew/php/codeception dependency: php56
Error: Cannot install homebrew/php/php56 because conflicting formulae are installed.`

Can it be dependent on changes in a03c9ebf5ae3157944a6668f499f923970d13044?

PR #2444, opened 2015-10-08
Add Puli formula

jenkoian:
Initially tried using the phar, but had some issues with checksums. Also, looking at https://github.com/Homebrew/homebrew-php/pull/1804 it seems tars are favoured anyway?


ablyler:
@jenkoian if there is a phar available, use that.  Look at this formula for an example: https://github.com/Homebrew/homebrew-php/blob/4e582a638e8ef18a885c4361d514655e81c61a74/Formula/behat.rb

jenkoian:
@ablyler changes made, I had originally done it as a phar but changed to a tar because of the PR I mentioned above. Cheers for heads up about the bottle thing, I had worked that out though, although only from reading through old PRs and references to the test bot. Documentation PR incoming! :)

PR #2453, opened 2015-10-12
Fix breaking change

michaeljs1990:
The default of 5.6 is failing if you have any other version installed. You should not have to install version 5.6 in order to install composer since it works with as low as php 5.3.


ablyler:
@michaeljs1990 the default should only be installed if none of the php versions are installed, including php53.  is this not working that way for you?

By the way I made this change to fix travis ci and brew bot issues.

BrianGilbert:
@ablyler I can confirm that it tries to install php56 even when php55 is installed

michaeljs1990:
Yeah, that is how i ran into this as well. I had php55 installed and it keeps failing saying that i need to unlink php55 first and then reinstall.

ablyler:
@michaeljs1990 i went ahead and made a different change that should fix this issue 113588a, thanks for the help

michaeljs1990:
awesome thanks for the fix.

PR #2482, opened 2015-10-26
Update arcanist.rb

michaeljs1990:
Remove sha. You can't use a sha when pulling master.


michaeljs1990:
Looks like your travis-ci builds don't handle head only formulas.

Issue #2523, opened 2015-11-17
Build without the legacy mysql extension (mysql_*)


schanzel:
It would be helpful if there was an option to omit the old and deprecated (as of 5.5) mysql extension.
```bash
brew install php56 --without-legacy-mysql
```
wich will skip
```ruby
args << "--with-mysql=#{HOMEBREW_PREFIX}"
```
and
```ruby
args << "--with-mysql=mysqlnd"
```
would be sufficient.

The use case for this change is using brewed php together with the AzizSaleh/mysql library (https://github.com/AzizSaleh/mysql).

PR #2524, opened 2015-11-17
Build without the legacy mysql extension (mysql_*)

schanzel:
Implemented as proposed in issue #2523


schanzel:
Alright, thank you!

PR #2542, opened 2015-11-27
Update php56-twig.rb to v1.23.1

leevigraham:


ablyler:
Thanks @leevigraham, but the sha256 was wrong. I've gone ahead and created PR #2545 which has the correct sha256.  Thanks!

PR #2546, opened 2015-11-29
Add formula for new mongodb driver

ablyler:
- Add description to legacy MongoDB driver formula


josegonzalez:
Can we add a warning to the post-install for the legacy drivers to push people to update their drivers?

ablyler:
@josegonzalez yeah, good idea. created issue #2547

PR #2548, opened 2015-11-29
Add missing descriptions to formula

alcaeus:
This PR adds a bunch of missing descriptions to formulae that were lacking them. In a few instances the package URLs were also out of date and have been corrected.

Some of the don't install the newest version of the respective package, this has not been changed in this PR.


alcaeus:
Well, looks like there are a few issues in the build - not sure this was a good idea after all. :disappointed:

ablyler:
@alcaeus not a bad idea at all! however, we might want to break this up into a few PRs. could create a script to create a PR for each formula.

alcaeus:
That sounds like an idea, I'll give that a shot.

alcaeus:
I created the first batch, the second batch will come as soon as GitHub unblocks my API tokens. It would seem that 46 pull requests is where it draws the line.

josegonzalez:
@alcaeus good work!

jenkoian:
Fair play! Was thinking it'd be cool if missing descriptions got caught by the travis build, and it seems that  brew audit does indeed support this https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/audit.rb#L378 it also seems that the brew test-bot should use strict also (if I'm reading the ruby correctly) https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/test-bot.rb#L568-L569 but looking at the build logs it doesn't seem to pass the strict param. So not sure, but would be cool to have this in the build to stop this happening again if someone can work out how :)

xu-cheng:
@ablyler @alcaeus Next time, please don't send massive PRs like this. Instead of please using one PR with multiple commits and cancel the CI build immediately. Otherwise, it would have drained almost our entire CI resources. Thanks.

alcaeus:
@xu-cheng Sorry about that :|

I was going to work on the failing builds first to see why they are failing (this is a good opportunity to catch any problems while installing packages), I can hold off on creating the remaining pull requests (there are 23 packages left, that would make it 23 pull requests). Just let me know how you want me to proceed.

PR #2621, opened 2015-12-03
Adds formula for phpunit 4.8.19

nickjanus:
Since some of us are stuck with php 5.5. :smile: Hopefully this is the right repo.


ablyler:
@nickjanus can you update the phpunit formula to print out an error if the current version of php is not supported?  and instruct them to install this new one?

javian:
@nickjanus ping ^^

nickjanus:
Whoops, closing this because I don't think this is all that useful.

PR #2660, opened 2015-12-07
Updated PHP extension "ev" to 1.0.0RC2

gcds:


ablyler:
@gcds thanks, but I rather not use an RC for anything except new versions (like php70).  I've gone ahead and used your new formula for PR #2662.  Thanks!

PR #2664, opened 2015-12-09
added php70-ssh2 extension

emir:
I couldn't see --with-ssh2 older php version implementations. But -because of dependent on ssh2- I added.

### DIFF
--- a/Formula/php70-ssh2.rb
+++ b/Formula/php70-ssh2.rb
@@ -1,0 +1,21 @@
+require File.expand_path("../../Abstract/abstract-php-extension", __FILE__)
+
+class Php70Ssh2 < AbstractPhp70Extension
+  init
+  desc "Provides bindings to the functions of libssh2 which implements the SSH2 protocol."
+  homepage "https://pecl.php.net/package/ssh2"
+  head "https://github.com/php/pecl-networking-ssh2.git", :branch => "master"
+
+  depends_on "libssh2"
+
+  def install
+    safe_phpize
+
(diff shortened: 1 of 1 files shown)


lucasmichot:
Thanks for submission @emir , can you check errors and update your PR ?

emir:
I checked but I couldn't understand why is it failing on "--with-ssh2" part. I saw almost the same usage in other extensions' implementations.

wizardfrag:
@emir --with-ssh2 requires an argument, which is the path to the ssh2 library. Without it it will use the standard ssh2 library. You can see this happening here: https://github.com/php/pecl-networking-ssh2/blob/master/config.m4#L7-L20

Perhaps this should be added as an argument to the formula, however I can't see many people having a different-than-system ssh2 installed?

If you remove that line, this formula will work as HEAD-only.

ablyler:
@emir can you fix these issues:

```
==> brew style php70-ssh2
== Formula/php70-ssh2.rb ==
C: 14: 48: Trailing whitespace detected.

1 file inspected, 1 offense detected

==> audit problems
homebrew/php/php70-ssh2:
 * Head-only (no stable download)
 * Description is too long. "name: desc" should be less than 80 characters.
Length is calculated as homebrew/php/php70-ssh2 + desc. (currently 106)
 * 14: Trailing whitespace was found

Error: 3 problems in 1 formula
```

ablyler:
I created PR #2898 to clean up the issue, thanks for the PR @emir !

PR #2671, opened 2015-12-12
mongodb 1.2.0-beta from PHP7 branch, head-only formula

risyasin:
Using PHP7 branch as head only formula.  I think will be useful for testing driver till it reaches a stable version.


alcaeus:
I don't think this should be merged without a big, fat beta disclaimer.

risyasin:
Totally agreed. Contains code has not released or **not even in master branch**. As i said this is for people who can not wait for a release or have a broken development environment after updating to PHP7 (like me)

alcaeus:
ext-mongodb 1.1.1 adds support for PHP 7, so I've created #2699 to update existing formulas and add php70.

PR #2673, opened 2015-12-13
Added missing package igbinary

aryeguy:


ablyler:
@aryeguy you can install igbinary via php56-igbinary.  Replace 56 with the version of your choosing.

PR #2700, opened 2015-12-23
Fixes libsodium dependency

cmckni3:


cmckni3:
Failing because it's using the bottled version.

Should I remove the bottle section from the formula?

cmckni3:
Thoughts @ablyler?

ablyler:
@cmckni3 what does changing this to a build only dependency fix? I think it needs to be a normal dependency, since it will link against it.

cmckni3:
@ablyler I see. You're absolutely right.

Is there any way to trigger a new bottle of the formula? The current bottle doesn't work on El Capitan.

cmckni3:
Closing. See #2715

PR #2719, opened 2016-01-05
Add php70-apcu-bc

mcfedr:
Since apcu v5 the backwards compatibility has been moved out into a separate module. This new formula installs this module.

The module includes the php functions `apc_*` (apcu now only provides `apcu_*`)

Also updated apcu such that apcu-bc can be built.

### DIFF
--- a/Formula/php70-apcu-bc.rb
+++ b/Formula/php70-apcu-bc.rb
@@ -1,0 +1,43 @@
+require File.expand_path("../../Abstract/abstract-php-extension", __FILE__)
+
+class Php70ApcuBc < AbstractPhp70Extension
+  init
+  desc "APC User Cache - BC"
+  homepage "https://pecl.php.net/package/apcu_bc"
+  url "https://pecl.php.net/get/apcu_bc-1.0.1.tgz"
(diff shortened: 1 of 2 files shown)


ablyler:
@mcfedr thanks, but let's wait until @krakjoe tags a release

mcfedr:
Updated with new tagged versions.

mcfedr:
@ablyler Do you have any pointers as to what I need to do to make the tests pass?

ablyler:
@mcfedr it looks like the one formula is trying to write to the php70 cellar.

mcfedr:
@ablyler It is, it installs the headers into the `includes/ext/apcu` folder. Then when apcu-bc is built it sees the headers there.

mcfedr:
@ablyler Finally got this building use the same technique as in other extensions with dependancies such as https://github.com/Homebrew/homebrew-php/blob/master/Formula/php56-http.rb

ahonymous:
Hi. It's grate PR. When was merged this?

AlexandreHagen:
Yes ! When it will merge? I need it.

PR #2792, opened 2016-01-17
Symfony installer 1.5.0

javian:


javian:
This is my first formula so any feedback is appreciated =) 

I don't quite get the functionality PhpMetaRequirement for instance. I looks like it checks if a brewed PHP is installed which is this case might not be required since the installer requires php to run although the built in OSX version would probably suffice since it will run on 5.4 and above.

The formula doesn't check for the PHP version (and there doesn't seem to be a standard way of doing this) but the software itself does.

javian:
This requires a php setting  phar.readonly=Off which is not set on the server. What would be the suggested way of dealing with this ?

mcfedr:
What if you just download the prebuilt phar? Is there any disadvantage to this?

javian:
The problem with it the prebuilt one is that it isn't versioned and that it is not uploaded to the release packages on github with a version number. Technically it against the core homebrew rule that they don't approve self-updating applications which the symfony installer (and several other tools in homebrew-php like composer and phpunit) but that rule doesn't seem to apply to the php tap. Both Composer and phpunit have a similar "download the latest version" url without a version number but also supply releases in their github repos.

javian:
I did submit a formula change to the box formula in #2802 that enables the required php setting for  the build so this might pass the tests now but I haven't resubmitted it yet due to lack of time but will try to do so over the weekend.

josegonzalez:
I'd rather we not continue adding non-versionable formula. This just means more maintenance hassle for installing packages.

I think composer and phpunit are outliers - most of the php community uses both, whereas there are many framework developers who *do not* use symfony.

josegonzalez:
Is there a reason the symfony community doesn't want to provide a versioned phar file to upload? I can't see a reason why they wouldn't, especially as they *do* create [releases](https://github.com/symfony/symfony-installer/releases).

javian:
They've been asked that question last year in symfony/symfony-installer#116 and said that they won't do it.

Formula/symfony-installer.rb
```diff
@@ -0,0 +1,27 @@
...
+  include Language::PHP::Composer
+
+  desc "Installer to create Symfony framework projects."
+  homepage "https://github.com/symfony/symfony-installer"
+  url "https://github.com/symfony/symfony-installer.git",
+    :tag => "v1.5.0",
+    :revision => "74311cd7ffacba6a1246d5cd5fc1a6d90037fe61"
+  head "https://github.com/symfony/symfony-installer.git"
+
+  depends_on PhpMetaRequirement
```
ablyler: Also add `depends_on PharBuildingRequirement`, `depends_on PharRequirement`, and the related `require` lines above for it.

ablyler:
I agree, we shouldn't add non-versioned formulas. However, it looks like the current version of this formula does indeed have a version, thanks to building the phar instead of using the pre-build one. I did add some comments for @javian, that will help clean this up a little bit.

Also, @javian can you update the commit message to prevent homebrew from spitting out these warnings?

```
Warning: Nonstandard bump subject: Symfony installer 1.5.0
Warning: Subject should be: symfony-installer 1.5.0 (new formula)
```

Thanks!

javian:
Thanks for the feedback and I can update the commit message no problem - where do you see that warning ?

I'm hesitant of adding the PharBuildingRequirement dependency since that will always fail unless you globally disable phar.readonly which was why I submitted #2802 for the 'box build' command specifically. For some reason the box build solution doesn't seem to be picked up by the CI environments. No other formulas seems to be using the phar building requirements so this would in that case be the first one.

ablyler:
I see the warning when running `brew pull https://github.com/Homebrew/homebrew-php/pull/2792`.

I'm getting an error when install on my box:
```
Andys-MacBook-Pro :: ~/Code/homebrew-php ‹master› » brew install symfony-installer
==> Installing symfony-installer from homebrew/php
==> Cloning https://github.com/symfony/symfony-installer.git
Updating /Library/Caches/Homebrew/symfony-installer--git
==> Checking out tag v1.5.0
==> composer --no-interaction install
==> box build
Last 15 lines from /Users/ablyler/Library/Logs/Homebrew/symfony-installer/02.box:
box
build

Building...



  [UnexpectedValueException]
  creating archive "/private/tmp/symfony-installer20160225-20764-6q9a7r/symfony.phar" disabled by the php.ini setting phar.readonly



build [-c|--configuration CONFIGURATION]



READ THIS: https://git.io/brew-troubleshooting
If reporting this issue please do so at (not Homebrew/homebrew):
  https://github.com/Homebrew/homebrew-php/issues
```

javian:
Weird since it works just fine on my setup but you show the same problem as the build environments. What does `php -i | grep phar` output on your machine ?

```
➜  ~  php -i | grep phar
Registered PHP Streams => https, ftps, compress.zlib, compress.bzip2, php, file, glob, data, http, ftp, phar, zip
Phar-based phar archives => enabled
Tar-based phar archives => enabled
ZIP-based phar archives => enabled
phar.cache_list => no value => no value
phar.readonly => On => On
phar.require_hash => On => On
➜  ~
```

javian:
I think I know what the problem is. After I committed #2802 the box bottle wasn't updated so the bottle contains the code which doesn't contain the wrapper change. Now that I've made a PR of the new box version in #2899 this should take care of itself.

PR #2822, opened 2016-01-29
remove php53-mustache as it no longer compiles

ablyler:


jbboehr:
It should still support PHP 5.3: https://travis-ci.org/jbboehr/php-mustache/jobs/101738322

PR #2828, opened 2016-02-04
PHP Code Sniffer (phpcs) & PHP Code Beautifier and Fixer (phpcbf)

DominoTree:
phpcs/phpcbf are nice to have and integrate with the PhpStorm IDE


DominoTree:
Looks like the Travis stuff failed due to a GitHub rate limit :P

ablyler:
Both of these are available via the `php-code-sniffer` formula.

DominoTree:
Thanks, I'd always thought it was odd that they didn't exist, guess I was looking for the wrong package name :D

PR #2863, opened 2016-02-05
php56-tidy: update bottle

ablyler:


DomT4:
Please take it a little easy with these tonight; the Mavericks bot is really struggling to keep pace lately and the queue gets very backed up.

PR #2871, opened 2016-02-11
add varnish extension for php70

razmo:
Add varnish extension for php 7.


ablyler:
@razmo it looks like this extension fails to compile against php7.

razmo:
Yes, it seem that there is a problem the extension :-(
You can close this PR if you want. I will make another when the extension is fixed

PR #2872, opened 2016-02-12
PR to add the latest DrupalConsole release.

darol100:
That is the latest version of DrupalConsole


ablyler:
@darol100 the sha256 hash needs to be updated as well

PR #2873, opened 2016-02-13
terminus 0.10.3 (new formula)

derimagia:
Adding new formula for Terminus: https://github.com/pantheon-systems/cli


Derimagia:
Error here is "Error: GitHub API rate limit exceeded for XXX.XX.XXX.XXX. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)". Any idea how we can resolve this?

Derimagia:
I made one small fix for testing, but I think the other issues have to deal with the github limit.

I am kind of confused at how the issue I just fixed didn't show up on my machine, something to deal with how system parameters are sent or something..

javian:
This has been causing problems in a number of cases. @homebrew/homebrew-php can you assist ?

javian:
I would also suggest that you use the release phar file instead so you don't have to compile it which makes the formula smaller and require less dependencies. Is there a reason why you can't use that ?

Derimagia:
I was copying what wp-cli and drush do, which are similar. I also wanted to in the future include bash/zsh autocompletion but was having issues (Unrelated, but see [here](https://github.com/pantheon-systems/terminus/issues/916))

Is this the way that's recommended by everyone? To switch to everything to use phars?

javian:
In that case you should leave it but as the Formula was constructed now there's really no need to not use the phar.

Derimagia:
Not sure who can answer this for me. Would really appreciate it!  (@ablyler ?)

ablyler:
@derimagia don't worry about TravisCI. That is still in testing. I'm reviewing this new formula now, thanks for the addition!

PR #2874, opened 2016-02-14
Adminer 4.2.4

javian:


javian:
It fails on

- ==> brew tests
- ERROR:  Could not find a valid gem 'bundler' (>= 0), here is why:
-           Unable to download data from https://rubygems.org/ - no such name (https://rubygems.org/latest_specs.4.8.gz)
- ERROR:  Possible alternatives: bundler
- Error: Failed to install/update the 'bundler' gem.
- ==> Installing or updating 'bundler' gem
- ==> FAILED

this looks like a travis problem.

PR #2889, opened 2016-02-18
Satisfy requirements if php formula is dependent

arnested:
Currently `PhpMetaRequirement` doesn't always behave as expected.

I.e. the [phan](https://github.com/Homebrew/homebrew-php/blob/master/Formula/phan.rb) formula specifies `php70` as dependency (and only works with PHP 7).

However if no PHP is yet installed `PhpMetaRequirement` will also add `php56` as a dependency because no linked formula exists.

The results is (as seen quite often with @BrewTestBot) that installation fails because `php56` will be installed and linked first. Then `php70` will fail to installed because `php56` needs to be unlinked first.

`PhpMetaRequirement` wouldn't have to fallback to the default formula if only it was aware that a PHP formula actually specified as a dependency.

This pull request implements that logic. Except Homebrew doesn't store all the collected dependencies anywhere. Below is a patch for [Homebrew core](https://github.com/Homebrew/homebrew) that stores all dependencies as a class variable in the `DependencyCollector` class.

Now for the discussions:

- Did I miss any edge cases that would still fail or make valid cases break?
- Is there a better way of achieving this? (preferably without needing changes to Homebrew core)
- If this is a valid plan should I/we try to get the patch below into Homebrew core?

```patch
diff --git a/Library/Homebrew/dependency_collector.rb b/Library/Homebrew/dependency_collector.rb
index a8e2609..f572a59 100644
--- a/Library/Homebrew/dependency_collector.rb
+++ b/Library/Homebrew/dependency_collector.rb
@@ -31,13 +31,19 @@ class DependencyCollector

   def initialize
     @deps = Dependencies.new
+    @@collected_dependencies ||= Set.new
     @requirements = Requirements.new
   end

+  def self.collected_dependencies
+    @@collected_dependencies ||= Set.new
+  end
+
   def add(spec)
     case dep = fetch(spec)
     when Dependency
       @deps << dep
+      @@collected_dependencies << dep
     when Requirement
       @requirements << dep
     end
```


arnested:
So... With `phan` as the example it actually only works with `brew install --build-from-source phan` and not with `brew install phan`.

The `satisfied?` method of `PhpMetaRequirement` is executed two times when building from source and only one time otherwise. It detects `php70` as a dependency each time it is executed though.

Well... Back to debugging. Comments still appreciated.

josegonzalez:
Slick. +1

arnested:
Ah!

Instead of checking the dependent PHP version in `satisfied?` it is better to override the `Requirement` class `to_dependency` method instead and returning a dependent PHP instead of the `default_formula`.

Now it works when installing both from bottles and source. Patch to Homebrew core still needed.

ablyler:
This is great @arnested!

ablyler:
@arnested did you create a pull request for core as well?

josegonzalez:
Note: we should wait a while for ^ to hit upstream, or at least log if the methods don't exist so that it doesn't break anyone's installation.

arnested:
I haven't created a pull request on core yet. I would like some feedback here before.

I'll create a pull request soon.

@josegonzalez, I did change the code so that will check the existence of the property on `DependencyCollector` before using it. Exactly to avoid breaking stuff where homebrew-php gets update before homebrew.

josegonzalez:
:+1:

ablyler:
@arnested this looks good to me, do you have any other changes you want to do before I merged it in?

arnested:
No, I have no other. But maybe we should wait to see the reaction on the pull request on core? Just in case they don't like the naming of the method (or the idea at all).

I'm just about to create the core pull request now (I just had to create some test coverage). But any minute now!

arnested:
Pull request created on core: Homebrew/homebrew#49361.

arnested:
The pull request to core has not been well received.

I am offline for the next 2.5 weeks so I won't be able to take it any further right now. Feel free to take this further if you want to.

javian:
Wouldn't it make more sense to construct a version check so you could specify a min/max version of php that a formula supports ? It wouldn't end up creating the dependency twice since it would depend on the right version to begin with.

ablyler:
@arnested / @javian / @josegonzalez please check out pull request #2896, it is a different approach that I think will yield the same request and doesn't require a change to homebrew core.

ablyler:
I went ahead and merged #2896

arnested:
I am back from vacation and happy to see you found a good solution! :+1:

PR #2892, opened 2016-02-21
Phalcon 2.0.10 for PHP 5.3

Zifius:


ablyler:
@Zifius it looks like this version is failing to compile against php53.

Zifius:
@ablyler yep, what to do in this case?

ablyler:
We can just leave it at the previous version.

PR #2896, opened 2016-02-24
Force PHP meta requirement to honor core php formula dependencies

ablyler:
When a core php formula is added as a dependency, the php meta requirement should honor it.


arnested:
Nice solution, @ablyler! :+1:

PR #2904, opened 2016-03-01
Add php7 event formula

dakota:


javian:
You need to remove the version from the formula to make it work

```
==> audit problems
homebrew/php/php70-event:
 * Stable: version 2.0.0 is redundant with version scanned from URL
```

ablyler:
@dakota can you make the change that @javian pointed out?

dakota:
I've removed the version from the formula.

PR #3068, opened 2016-04-04
Update php70-msgpack.rb

joanmon:
Php7 branch seems no longer available. It was merged into master.


ablyler:
Does 2.0.1 support PHP7?

javian:
According to this  https://github.com/msgpack/msgpack-php/commit/b29f3fd572b9d2ad8c053efa6279f93862f63657 it seems that 2.0 and onwards supports php7 so 2.0.1 should be fine.

javian:
@joanmon can you update the PR to use versions to complement the head formula or replace it ?

joanmon:
Do you mean to be able to select specific tags or master by default?

javian:
Not quite - look at https://github.com/Homebrew/homebrew-php/blob/master/Formula/php56-msgpack.rb for instance and you will see that it uses a specific release not using git so something more along those lines.

joanmon:
@javian Could you check it? I don't know if It is correct. Thank you

javian:
Looks great but lets just wait and see if it passes the travis tests. =) Thanks for the PR.

PR #3082, opened 2016-04-09
drupal-code-sniffer 8.x-2.7

arnested:


javian:
```
==> Using the sandbox
/usr/bin/sandbox-exec -f /tmp/homebrew20160409-8368-1wpvuwj.sb nice /System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/bin/ruby -W0 -I /usr/local/Library/Homebrew -- /usr/local/Library/Homebrew/build.rb /usr/local/Library/Taps/homebrew/homebrew-php/Formula/drupal-code-sniffer.rb --build-bottle --verbose
==> Downloading http://ftp.drupal.org/files/projects/coder-8.x-2.7.tar.gz
Already downloaded: /Library/Caches/Homebrew/drupal-code-sniffer-8.x-2.7.tar.gz
==> Verifying drupal-code-sniffer-8.x-2.7.tar.gz checksum
tar xf /Library/Caches/Homebrew/drupal-code-sniffer-8.x-2.7.tar.gz
Error: Operation not permitted - /usr/local/share/drush
```
Should there not be a dependency to the drush formula in the drupal-code-sniffer since its using drush ?

arnested:
Yes, there should be a dependency on `drush`. Added and pushed.

arnested:
But the missing dependency was actually not why the test-bot failed. It failed because it tried to link the standard into the shared drush folder while building the bottle. Splitting the `install` method into an `install` method and a `postinstall` method did the trick (and is the right behaviour).

PR #3088, opened 2016-04-10
phplint 3.0 20160307

javian:


javian:
Mavericks comes with PHP 5.4 so I'm assuming it fails on that platform for that reason and there's no dependency on PHP in the formula. Should we add a dependency to PHP on Mavericks only ?

ablyler:
@javian that sounds like a good idea to me

Issue #2544, opened 2015-11-27
php**-intl broken after ﻿﻿icu4c upgrade


derrabus:
I've upgraded my homebrew packages today and got an update for icu4c from 55.1 to 56.1. After that upgrade, php fails to load its intl extension:

```
PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/local/opt/php70-intl/intl.so' - dlopen(/usr/local/opt/php70-intl/intl.so, 9): Library not loaded: /usr/local/opt/icu4c/lib/libicui18n.55.dylib
  Referenced from: /usr/local/opt/php70-intl/intl.so
  Reason: image not found in Unknown on line 0
```

The output above is from php 7.0 RC7, but I can also reproduce this with the corresponding php 5.6 and 5.3 packages. I assume that 5.4 and 5.5 are also affected, but I don't have those packages installed at the moment. Running `brew reinstall php70-intl` does not fix the problem.

```
$ brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: You have unlinked kegs in your Cellar
Leaving kegs unlinked can lead to build-trouble and cause brews that depend on
those kegs to fail to run properly once built. Run `brew link` on these:
    php53
    php56
```

derrabus:
If I reinstall the package from source, the errors are gone:
```bash
brew reinstall --build-from-source php70-intl
```

timwangdev:
Same issues here.
```
Warning: PHP Startup: Unable to load dynamic library '/usr/local/opt/php70-intl/intl.so' - dlopen(/usr/local/opt/php70-intl/intl.so, 9): Library not loaded: /usr/local/opt/icu4c/lib/libicui18n.55.dylib
  Referenced from: /usr/local/opt/php70-intl/intl.so
  Reason: image not found in Unknown on line 0
```

DomT4:
[Self-reference](https://github.com/Homebrew/homebrew-php/issues/1701#issuecomment-160036522): Someone needs to submit a PR to bump the revisions of the formulae involved and then merge that PR with the revision bump and new bottles.

derrabus:
php56-intl and php70-intl have been fixed with upstream version bumps (5.6.16 and 7.0 RC8). Other php branches (5.3, 5.4, 5.5) might still be broken.

bskendig:
[...]

PR #3090, opened 2016-04-11
revisions for icu4c

skystrife:
icu4c got a version bump (see Homebrew/homebrew-core#183).

This PR includes revision bumps for php packages that depend on ICU.

### DIFF
--- a/Formula/php53-intl.rb
+++ b/Formula/php53-intl.rb
@@ -7,6 +7,7 @@ class Php53Intl < AbstractPhp53Extension
   url PHP_SRC_TARBALL
   sha256 PHP_CHECKSUM[:sha256]
   version PHP_VERSION
+  revision 1
 
   bottle do
     revision 1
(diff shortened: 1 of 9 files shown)


javian:
This should also resolve #2544 if its still an issue.

craigcarnell:
Also get the same error with hhvm

craigcarnell:
please merge asap

javian:
We will get to it when we can. I have the access to do it but I don't know the process (yet) so I'm awaiting some assistance from fellow repo maintainers to help explain how it works.

jpruizs:
Hi,
I am running with this issue and not sure how to solve it. I've reinstalled php56-intl and built from source but still happening. This is the current error:

> PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/extensions/no-debug-non-zts-20121212/intl.so' - dlopen(/usr/lib/php/extensions/no-debug-non-zts-20121212/intl.so, 9): Library not loaded: /usr/local/opt/icu4c/lib/libicui18n.56.dylib\n  Referenced from: /usr/lib/php/extensions/no-debug-non-zts-20121212/intl.so\n  Reason: image not found in Unknown on line 0

Is there anything else I could do to fix it?
I can see this file `libicui18n.57.dylib` but not this one `libicui18n.56.dylib`

Thanks

PR #3100, opened 2016-04-13
add GitHub issue template

apjanke:
Hi folks,

GitHub added support for issue and PR templates recently. So now, instead of having your README tell people about the issue format and provide some Markdown to copy and paste, you can use the template to automatically pre-populate all new issues with that template. Homebrew Core is doing this already for theirs; go to https://github.com/homebrew/homebrew-core/issues and click "New Issue" to see it in action.

This PR changes the homebrew-php repo to do the same. I modified the template a little bit. This format should work better when people do an actual copy and paste, because the `gcc -v` and `php -v` output is multiple lines, and won't fit in a single Markdown table cell (without munging to use `<br>`). And it recommends use of `brew gist-logs` instead of manual gist creation. `brew gist-logs` includes `brew doctor` output, along with more detailed build logs.

If you wanted, you could even add your own command, like `brew php-bug-info` that would programmatically query and fill out all the values in this bug template so the user wouldn't have to run them and do multiple copy/pastes.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,20 @@
+## Description
+
+_Problem description goes here_
+
+## Context and Logs
+
+Parameter | Value
+--------- | ------------------
+**OS X Version:**       | _your OS X version_
+**Homebrew Version:**   | _your Homebrew version_
+**PHP Version in use:** | _your PHP version_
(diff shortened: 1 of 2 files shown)


josegonzalez:
Good call. @ablyler anything you think we should add to this?

ablyler:
Looks good to me, I'll merge it in! Thanks @apjanke

Issue #3097, opened 2016-04-12
I can't make php55 connect to IBM DB2 server because of unixodbc dependency (El Capitan)


mikekamornikov:
It works just fine with default osx php:
```txt
PHP 5.5.31 (cli) (built: Feb 20 2016 20:33:10)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2015 Zend Technologies
```
```sh
Configure Command =>  '/Library/Caches/com.apple.xbs/Binaries/apache_mod_php/apache_mod_php-102.1~27/TempContent/Objects/php/configure'  
'--prefix=/usr' '--mandir=/usr/share/man' '--infodir=/usr/share/info' 
'--disable-dependency-tracking' '--sysconfdir=/private/etc' 
'--with-libdir=lib' '--enable-cli' 
'--with-iconv=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.Internal.sdk/usr' 
'--with-config-file-path=/etc' 
'--with-config-file-scan-dir=/Library/Server/Web/Config/php' 
'--with-libxml-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.Internal.sdk/usr' 
'--with-openssl=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.Internal.sdk/usr/local
```

And does absolutely nothing with brew version:
```sh
Configure Command =>  './configure'  '--prefix=/usr/local/Cellar/php55/5.5.34' 
'--localstatedir=/usr/local/var' '--sysconfdir=/usr/local/etc/php/5.5' 
'--with-config-file-path=/usr/local/etc/php/5.5' '--with-config-file-scan-dir=/usr/local/etc/php/5.5/conf.d' 
'--mandir=/usr/local/Cellar/php55/5.5.34/share/man' '--enable-bcmath' '--enable-calendar' 
'--enable-dba' '--enable-exif' '--enable-ftp' '--enable-gd-native-ttf' '--enable-mbregex' '--enable-mbstring' 
'--enable-shmop' '--enable-soap' '--enable-sockets' '--enable-sysvmsg' '--enable-sysvsem' 
[...]

PR #3102, opened 2016-04-14
Add --without-unixodbc

mikekamornikov:
see https://github.com/Homebrew/homebrew-php/issues/3097
- [x] make `unixODBC` dependency configurable
- [x] don't add `--with-pdo-odbc=unixODBC` and `--with-unixODBC` (in case of `--without-unixodbc`)
- [x] tests


javian:
Thanks @mikekamornikov !

PR #3120, opened 2016-04-28
Adds a new php70-gearman formula that updates php56-gearman

stephenorr:
This relies on wcgallego/pecl-gearman.

This is because PECL doesn't yet contain a Gearman extension for PHP 7; all efforts to this
end are being made by the community - which is fine, but makes life difficult for those of us
who want to use PHP 7 in the meantime.

Since this PR was originally created, wcgallego/pecl-gearman now has a stable tag for PHP 7; I've switched the formula to use that instead of the master branch I was using previously.


trickreich:
Add this please

javian:
THanks for the update @trickreich - you can remove the version line because it should be able to figure that out on its own now that the version is part of the release.

stephenorr:
Closing this one, because it didn't quite work; I've since gotten it to pass `brew audit` on a different branch so will submit a new PR from there.

PR #3159, opened 2016-05-06
tokyo-tyrant: migrate from core

apjanke:
Hi, Homebrew PHP folks,

We were planning on boneyarding `tokyo-tyrant` because it doesn't meet core Homebrew's maintenance standards. (https://github.com/Homebrew/homebrew-core/pull/816) There are some PHP formulae that depend on it. Would you like us to just move `tokyo-tyrant` over here, so you don't have to remove the dependencies for it? Since this is a repo with versioned formula and longer lives, it might still be acceptable under your guidelines.

### DIFF
--- a/Formula/tokyo-tyrant.rb
+++ b/Formula/tokyo-tyrant.rb
@@ -1,0 +1,39 @@
+class TokyoTyrant < Formula
+  desc "Lightweight database server"
+  homepage "http://fallabs.com/tokyotyrant/"
+  url "http://fallabs.com/tokyotyrant/tokyotyrant-1.1.41.tar.gz"
+  sha256 "42af70fb9f2795d4e05c3e37941ce392a9eaafc991e230c48115370f6d64b88f"
+  revision 1
+
+  bottle do
+    cellar :any
+    revision 1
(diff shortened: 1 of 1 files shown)


ablyler:
I vote for removal, what does everyone else think? /cc @Homebrew/php

javian:
+1 for removing it

apjanke:
```
$ brew uses tokyo-tyrant
homebrew/php/php53-tokyotyrant                            homebrew/php/php55-tokyotyrant
homebrew/php/php54-tokyotyrant                            homebrew/php/php56-tokyotyrant
$
```

I'm not a PHP user or expert, but: it looks like these packages are just bindings or wrappers for `tokyo-tyrant`, so removing it just means removing these packages as well, right?

ablyler:
@apjanke yep, we just need to remove those formulas

apjanke:
Okay. I'll add a PR for it.

What's the procedure for removing formulas from homebrew-php? Do you folks migrate them to homebrew-boneyard too, or just delete them?

apjanke:
Here's the replacement PR to delete them: https://github.com/Homebrew/homebrew-php/pull/3165

javian:
Closing this PR since it has been superseded.

PR #3181, opened 2016-05-21
Update to mailparse-3.0.1 for php70-mailparse

mta59066:
mailparse-3.0.0 has a segfault issue on php70


javian:
Thanks for the PR @mta59066

PR #3272, opened 2016-06-17
Updated php56-protobuf version.

viniciusjarina:

---


javian:
I think we have a problem with this formula due to how it uses the version. I believe the system will see this as an older version going from 5 to1 so the upgrade functionality won't work.

viniciusjarina:
Well the previous formula was already using the hash as version, the brew upgrade here locally without any problem. I don't think it will check the lexicographical difference of the versions

viniciusjarina:
@javian ^^

javian:
According to @DomT4 "the version is implicit or explicit it determines update. If you do something like updating a formula from version `1.0` to version `0.1` it won’t update for users." - so according to that it would not work since the version you have submitted is "lower" than the one in the currently formula. 

Also, you don't have to remove the bottle code since the bot will take care of that itself.

DomT4:
You can do a manual comparison fairly easily FWIW, if you drop into `brew irb` and then run something like:

```ruby
Version.new("1fa365d") > Version.new("5839c25")
=> false
```
Looks like @javian is correct here.

viniciusjarina:
Ok 😄 

But since the previous version was already using the git hash as version what you guys suggest?

Bump version to `5839c25.2` ? or bump to `5839c26` ?

DomT4:
Yeah, this is definitely one of the reasons Homebrew core doesn't accept untagged/unversioned releases 😅. Maybe something semi-sane would be to do:

```
version "62016613"
```
Where the `6` makes it bigger than the `5` of the existing version number, and then everything after is the date the hash was committed on. That should mean going forwards everyone gets updates & you don't fall foul of the versioning logic, providing you don't go backwards in time of course.

viniciusjarina:
Ok. Updated here https://github.com/Homebrew/homebrew-php/pull/3272/commits/8690bffbfac9f53e8114efe446ce5e9a641967b6 

Thank you.

Formula/php56-protobuf.rb
```diff
@@ -5,18 +5,14 @@ class Php56Protobuf < AbstractPhp56Extension
...
-    sha256 "043dccfb1da3e3f9cf303511d19e96ee9dfbbfa4cd8c25018f2ef5356073557c" => :el_capitan
-    sha256 "633d0171ed34a8e4cd15d7f5a756faf924f5dac0178807d6d5e32c92c582bf29" => :yosemite
-    sha256 "fdf187e3c88665af8d92a290ad3ca4682d9a0d5e41d1f52333eb9baff7c570ef" => :mavericks
-  end
-
+  url "https://github.com/allegro/php-protobuf/archive/1fa365d6b29645d9585019b77600126c896c4cf6.tar.gz"
+  sha256 "383d69a5c630b0f134ab571f182848145a108695d3425af2988cd0d16402cac2"
+  version "620160615"
+  # Note: This version is not the actual version of https://github.com/allegro/php-protobuf
+  # Since the previous version of the formula was using the git hash, in order
```
DomT4: Looks like there's some whitespacing on this line:
```
* 12: Trailing whitespace was found
```

javian:
Do we really need the 6 ?
```
2.0.0-p648 :010 > Version.new("5") > Version.new("20160101")
 => false
2.0.0-p648 :012 > Version.new("5") > Version.new("2")
 => true
2.0.0-p648 :013 > Version.new("5") > Version.new("20")
 => false
```
Just putting the date in should create a larger version number and would look neater don't you think ?

viniciusjarina:
@javian removed

DomT4:
> Do we really need the 6 ?

Apparently not. I thought due to the way `brew` calculates versions it may be necessary, but seems fine:
```ruby
irb(main):002:0> Version.new("20160615") > Version.new("5839c25")
=> true
```

javian:
After giving this some more though I think the better way to solve this now is to use the following notation.

`<current version>.<date of commit from branch>`

so in this case it would be

`1.1.0.20160101` if the relaese date was first on January 2016.

I do see now however that a new version has been released so we can put that in instead but if this situation would occur again I think we have a good way of solving it. Can you please update the formula for 0.11.0 ?

This might still cause issues for people who are using the current version but now that there's an official release lets revert to using that and make it right from now on.

javian:
PIng @viniciusjarina ^^

viniciusjarina:
OK . I will bump to the latest version fix the version.

javian:
Fantastic!

javian:
Ping @viniciusjarina ^^

Formula/php56-protobuf.rb
```diff
@@ -5,18 +5,9 @@ class Php56Protobuf < AbstractPhp56Extension
...
-    cellar :any_skip_relocation
-    revision 1
-    sha256 "043dccfb1da3e3f9cf303511d19e96ee9dfbbfa4cd8c25018f2ef5356073557c" => :el_capitan
-    sha256 "633d0171ed34a8e4cd15d7f5a756faf924f5dac0178807d6d5e32c92c582bf29" => :yosemite
-    sha256 "fdf187e3c88665af8d92a290ad3ca4682d9a0d5e41d1f52333eb9baff7c570ef" => :mavericks
-  end
-
+  url "https://github.com/allegro/php-protobuf/archive/v0.11.1.tar.gz"
+  sha256 "949ece1ded96ff19597cd1f8ab5febffa77bcdf30c679d3f440fb42e8a090431"
+  version "0.11.1.20160829"
```
javian: You can remove this line and let homebrew manage the versions.

viniciusjarina: @javian but you asked to do the version this way https://github.com/Homebrew/homebrew-php/pull/3272#issuecomment-237477976 😜

javian: Yes but that was for the non versioned released so that's out the window now =)

DomT4: You'll need to use the new `version_scheme` method if you want users automatically updated. At the moment the existing version is considered newer than the `0.11.1` replacement.

javian: Excellent! That's going to solve a lot of issues ahead - @viniciusjarina you can see an example in this PR  https://github.com/Homebrew/homebrew-core/pull/4006/files .

DomT4: Yeah, thankful we finally got it done. If anyone's familiar with Debian it's basically an epoch.

javian:
@viniciusjarina one more change and we're good to go!

javian:
@viniciusjarina ping ^^

Formula/php56-protobuf.rb
javian: @viniciusjarina will you make this change ? Otherwise I'll close the issue in 2 weeks.

viniciusjarina:
Hi @javian I am really sorry.

I am no longer using php-protobuf. I think you can close it ;)

Thank you. and Sorry

Issue #3271, opened 2016-06-16
What is PHP 7.1 support roadmap


pinepain:
Subj. What is the roadmap for PHP 7.1 support?

javian:
I don't think anyone is working on it yet. Now that the 7.1 alpha is out we could do a release version based on that and looking at the changes I couldn't see anything major that changed that would affect using what's already there for 7.0. If you would be interested to submit a PR then you are more than welcome to do so.

pinepain:
Thanks. I also can't recall any major or internals changes that will break extension or requires some special handling. 

If no one already on it, I will do some more testing locally and provide a PR.

javian:
Thank you that would be much appreciated!

pinepain:
PR https://github.com/Homebrew/homebrew-php/pull/3273 submitted. PTAL & try someone.

pinepain:
PR https://github.com/Homebrew/homebrew-php/pull/3276 with formulae submitted.

javian:
Closing this now since support for 7.1 has been merged and actively monitored.

PR #3274, opened 2016-06-18
Add PHP 7.1.0-alpha.1 core formula (php71)

pinepain:

-----

As per https://github.com/Homebrew/homebrew-php/issues/3271.

Known issues with PHP 7.1.0 alpha1:

 - xdebug won't compiled due to bug with `php_log_err macro` https://bugs.xdebug.org/view.php?id=1314#c3635 (already fixed in https://github.com/php/php-src/commit/d2bc82586b67882fdcdabac8e05cfbb4104bfd47).

 While this is already fixed less likely we need to patch php ourselves for that, based on [7.1 release schedule](https://wiki.php.net/todo/php71) next alpha will be available on Jun 23 2016, so we can ignore this bug or wait for alpha2 or apply that fix in fork or beg for alpha1.5.


javian:
As far as the header checks go I can see the same error coming up in `brew audit --strict --online php70` so I wouldn't worry about that for now.

pinepain:
Yeah, that's expected. Should formulae goes within the same PR? 

After I started to build formulae, I had to patch php itself to have xdebug and use some non-tagged versions, like for php-ast. While it's ok for alpha version, I doubt to add them into core tap. On the other hand having and maintaining own tap, e.g. pinepain/php71 may limit php71 adoption. Any suggestions?

javian:
Lets keep the core formula seperate and it can be merged. As far as the extensions and modules go I think we should only add the formulas the work and exclude xdebug for instance until there's an officially released version that properly supports it - it is alpha software after all =).

pinepain:
Yeah, that's good idea to merge just core formula first. Let's do it.
 
-----
All below is for the future PR with 7.1 formulae.

Xdebug itself has php 7.1 support, that's solely `php.h` header problem (see description in PR message). In my github fork I applied fix to php 7.1 branch and rolled custom alpha 1.5 release. Basically in my tap it's known to work (see https://github.com/pinepain/homebrew-php fork). Another issue is that php-ast version that adds 7.1 support not yet tagged to I referred to tag. Some other packages like phan depends on it. Once again, for alpha-quality software it's fine, imo. And without xdebug running on alpha version might not be a reasonable solution, personally in my tasks, developing without xdebug not a good idea. In case we decide to go with my custom php7.1 alpha1.5 release (https://github.com/pinepain/php-src/releases/tag/php-7.1.0alpha1.5) that can be done in different PR or we can delay release to wait for alpha2 which should be on Jun 23 2016.

javian:
Thanks for the PR @pinepain - I think it would be best to wait for the official release so its more manageable and we don't mix in too much custom changes.

PR #3276, opened 2016-06-21
Php71 formulae

pinepain:

-----
This PR is complementary for  #3274. As described and discussed in it, xdebug not included. We'll add it when PHP 7.1.0-alpha2 will be available (apr. Jun 23 2016).


javian:
I don't think that this is going to work because it breaks the build systems. We would need to add all formulas as separate PRs to test them. This is how it has been done before.

pinepain:
I see. Will do it separately. Should I also update php70 or php5x formulae in case of formula version upgrade?

pinepain:
And one more question, can we build phpdbg by default (e.g. remove with-phpdbg in a favor of without-phpdbg). For example, on ubuntu php is built with phpdbg and it kinda handy to have it. What do you think (sure, as a separate PR).

javian:
If you can then please submit the updated versions for the modules for 7.0 and 5.x. For the phpdbg change I would suggest you make a pr and I can ask the others to give their input on the change. At first glance it doesn't seem to be a big change and assuming we revert the option to "without phpdbg" then you can always opt out. I'll also check if there are some stats on the option usage for php to see if we have some numbers to support the pr.

pinepain:
Will do it in separate pr. Should we add

PR #3278, opened 2016-06-22
Enable phpdbg by default

pinepain:

-----

This PR introduce phpdbg building by default for all supported php version (php54, php56, php56, php70 and php71).


pinepain:
Another option is to put `without-phpdbg` option in abstract formula so all php version (except 5.3) will have it which I like better while it's cleaner. The only thing I'm not sure how to handle proper options rename, it doesn't looks like `deprecated_option` can be used in a situation when we flip option.  Or it's fine to break options BC? Suggestions?

pinepain:
@javian can you give it a look?

javian:
I don't see how this can work. Is it possible to not build phpdbg for 7.0 and 7.1 for instance ? The code doesn't add up to me but I might be blind =)

pinepain:
@javian `--without-phpdbg`? I think we can even build phpdbg by default for all versions that support it.

javian:
@BrewTestBot test this please

pinepain:
@javian what do you think about removing `with-phpdbg` in a favor of `without-phpdbg`? This will let us keep phpdbg flags handling in abstract formula, which reduces code duplication.

javian:
@pinepain I think that sounds better in case it causes problems or people would want to opt out of using it.

pinepain:
What can be a reason for opt-out in your opinion? For me the only reason for that is if one need to use dtrace. But even after adding phpdbg by default our build will be even closer what debian/ubuntu provides, which sounds good for me.

javian:
I'm starting to think this could be a good idea but we should move all of the code to abstract-php and not keep in the seperate phpXX formula files to make it easier to maintain.

pinepain:
Exactly. I will spent time on that later this week, I already have it done to some degree locally, but want to make sure it works one more time.

pinepain:
@javian PR updated, ptal.

Formula/php54.rb
```diff
@@ -9,14 +9,10 @@ class Php54 < AbstractPhp
   url PHP_SRC_TARBALL
   sha256 PHP_CHECKSUM[:sha256]
   version PHP_VERSION
-  revision 4
```
javian: The revision lines needs to not be removed and increased by 1.

pinepain: Should it be moved then to `bottle do` section?

pinepain: I don't quite get how revision mechanism work (https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Formula-Cookbook.md#formulae-revisions) and what's the difference (if any) between `revision` in `bottle do` block and in formula, can you point me to some doc/source code or provide brief explanation, please?

javian:
Not that I've seen. I've only gotten bits a pieces of it but as far as I've understood it the following applies

1. Never touch the bottle block, the bot does that for us
2. Revision always goes outside the bottle block when its manually added

As far as I know the revision statement can only exist once either in or outside the bottle code.
It should all originate from the Formula module https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb

Formula/php54.rb
javian: Not that I've seen. I've only gotten bits a pieces of it but as far as I've understood it the following applies

Never touch the bottle block, the bot does that for us
Revision always goes outside the bottle block when its manually added
As far as I know the revision statement can only exist once either in or outside the bottle code.
It should all originate from the Formula module https://github.com/Homebrew/brew/blob/master/Library/Homebrew/formula.rb

javian: so don't move it to the bottle section, keep it and increment.

pinepain: Ah! I found it:

[class-level revision](https://github.com/Homebrew/brew/blob/b8ce1fe1b2abfc6b38a6c6451ef8074f2a83a8eb/Library/Homebrew/formula.rb#L112):

> ```
>   # Used for creating new Homebrew versions of software without new upstream
>   # versions.
>   # @see .revision
> attr_reader :revision
> 
> ```

[bottle-leverel revision](https://github.com/Homebrew/brew/blob/b8ce1fe1b2abfc6b38a6c6451ef8074f2a83a8eb/Library/Homebrew/formula.rb#L1770):
>     bottle do
>     ...
>       revision 1 # Making the old bottle outdated without bumping the version/revision of the formula.
>     ...
>     end</pre>

I will fix PR according to this info.

pinepain:
Updated one more time with leaving bottle hashes as is (with some re-arrange for consistance) and increment formula revision by 1.

javian:
Thanks @pinepain - since there seems to be a number of issues with the php54 which breaks the test bot and the building of the bottle I'm thinking we need to split the code from updating the core php Formulas revision so that we don't build new bottles and the put in another PR which updates the bottles for each individual php version.

javian:
https://github.com/Homebrew/homebrew-php/issues/3460 - this seems to be a blocker in php54 which currently breaks the build.

pinepain:
Tried to re-run after rebase, no luck.

Also, this error message:

https://bot.brew.sh/job/Homebrew%20PHP%20Pull%20Requests/2069/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/uses_homebrew_php_php70/
```
        Error: /usr/local/Library/Taps/homebrew/homebrew-core/Formula/isync.rb:9: syntax error, unexpected <<, expecting keyword_end
<<<<<<< Updated upstream
  ^
/usr/local/Library/Taps/homebrew/homebrew-core/Formula/isync.rb:14: syntax error, unexpected ===, expecting keyword_end
=======
   ^
/usr/local/Library/Taps/homebrew/homebrew-core/Formula/isync.rb:16: syntax error, unexpected >>, expecting keyword_end
>>>>>>> Stashed changes
```
seems weird to me.

Will now try without touching formula revision.

javian:
There was an update to isync.rb yesterday but it seems to be ok so maybe it was just unfortunate timing.

javian:
@BrewTestBot test this please

javian:
The isync problem was fixed on the build bot so lets give this another go.

javian:
@BrewTestBot test this please

pinepain:
@javian I just triggered a build =)

javian:
@pinepain =)

pinepain:
All builds are red :(. Is php54 a blocker or such type of changes leads to massive build when all formulas got rebuilt and CI can't handle that? Or something wrong with changes itself?

javian:
There are several things going on here. The isync issue is an upstream brew issue the some of the php54 extensions seems broken or don't work well together and it looks like the test bot doesn't honor the conflicting formulas (I'll file that with brew) so lets see if we can make this run green eventually =) I've also put in a PR to fix the dependency issue with dbus so that issue should be sorted once that is merged.

pinepain:
@javian it looks like apcu-bc(loaded as apc) depends on apcu (loaded as apcu) (https://github.com/krakjoe/apcu-bc/blob/master/config.m4#L9), but php54 should not be affected as it uses apcu.

As to that isyn issue (e.g. [this report](https://bot.brew.sh/job/Homebrew%20PHP%20Pull%20Requests/2081/version=el_capitan/testReport/junit/brew-test-bot/el_capitan/uses_homebrew_php_php54_dbus/)), it looks like formula contains merge conflicts, isn't it?

javian:
You see some details from the brew devs in my report here https://github.com/Homebrew/brew/issues/661  and yes you are right about apcu-bc but that an option that needs to be turned on in the [apcu Formula](https://github.com/Homebrew/homebrew-php/blob/master/Formula/php54-apcu.rb) so that's not the default.

pinepain:
Still failing :(. I see in travis same problem with apc already loaded.

javian:
I found the problem with APC. Unlinking the formula does not disable it since the config points to `/usr/local/opt/` which is not affected by `brew unlink` - this is intentional behaviour. Trying to figure out if there's an easy way to get around this.

pinepain:
@javian Any news related to this issue? Are there anything I can help with?

javian:
@pinepain not yet - it turns out to be more tricky than I would have thought to have the tests pass so its not an fix to make it work and it would to change how extensions are installed and managed. Maybe its not even feasible. Btw, I checked the ubuntu build logs and they do not build phpdbg by default as far as I could tell - did you have a reference to that ?

pinepain:
@javian it looks like my bad - phpdbg in debian/ubuntu is built only on demand - https://anonscm.debian.org/cgit/pkg-php/php.git/tree/debian/rules#n242, sorry for misleading you. I had phpdbg always installed by provisioner, so somehow missed tha  

So, is having it pre-build may still be a good idea, right? Considering that current xdebug doesn't support php71, it still arguable whether to have phpdbg in base package is nice to have or not.

Btw, Is current issue will block us from performing php versions upgrade?

pinepain:
At this point I decided to revoke this PR based on discussion above and while it is come up that phpdbg is not served by default even in debian/ubuntu and that was just my local build with phpdbg by default.

I will provide cleanup of phpdbg conditions (they can be unified and moved to abstract formula) in a separate PR.

javian:
Maybe there's and argument for making it a seperate formula ? Like php70-phpdbg. That would match what the linux distros are doing and that might work for homebrew as well don't you think ?

pinepain:
@javian exactly. I will play with it, but not in this PR.

PR #3331, opened 2016-07-03
Added pdo-firebird for PHP7.0

martimarkov:
The template used for the formula was `php70-pdo-pgsql.rb`


martimarkov:
The built failed because of GitHub api restrictions.

javian:
Since this is missing the Firebird headers then it can't be built and therefore it can't be included for the time being. As you point out a Formula for Firebird needs to be added before this can go anywhere. Closing this RP.

PR #3348, opened 2016-07-10
Add dev tools requirement in install doc

gagarine:

-----

Command line developer tools seem needed to install PHP formula. So this should be in the install section, not only in the troubleshooting section. I think this can avoid a lot of confusion and bad bug report.

I was even thinking taking the install section on the top. But as this is another problem I will split it in an other pull request.


ablyler:
@gagarine we didn't have it in the install documents, because it is part of the homebrew install documents.

gagarine:
Right. but sometimes.. it's help to duplicate information and giving it to the peoples before they actually fail into the trap.

Specially when tools absolutely want one liner install instruction:
screen shot 2016-07-21 at 21 30 10

But I don't care, I'm not the one maintaining this project.

PR #3399, opened 2016-07-27
travis.yml: tweak settings.

MikeMcQuaid:
General cleanup and use only one OS X worker (as these workers are shared between the Homebrew organisation and Jenkins CI does the bottling for you).


javian:
Thanks @MikeMcQuaid !

PR #3402, opened 2016-07-28
php54 bottle update

javian:

-----


javian:
There's something very odd with this bottle update in testing. The results looks similar from when the bottle last was updated last year so I will let it pass this time and see if we can get it resolved somehow. It seems to be failures on apc and couchbase.

PR #3414, opened 2016-07-28
Add php7x-uv formula

pinepain:

---


pinepain:
This is head-only formula while there are no tagged version for php7x in a source repo, so it expected to fail. Are there any way to let bot know that this is head-only formula and it should be built with `--HEAD` option?

javian:
No there's no way of telling the bot that as far as I know. This plugin is experimental so is there a good argument why we should include it in homebrew-php ?

pinepain:
There is nice async framework - [amphp/aerys](https://github.com/amphp/aerys) which can use various async io backends, including uv, which shows the best performance. While it is still very experimental, I guess it would be nice to help to spread it usage among users. It will also allow to keep consistence between local dev env and production for those who are on macs.

javian:
I just noticed that if you have installed a HEAD formula it won't upgrade to a regular version if one is provided (php70-redis was a head only formula until now). The regular homebrew upgrade process doesn't tell you at all but if you trigger the upgrade manually it will complain and you can deal with it by removing and reinstalling the module. This might be an OK behaviour if you do install formulas with `--HEAD` .

pinepain:
Interesting. I asked Bob Weinand to tag some version, but from our last talk he planned to do some refactoring (it's kinda constant process to want to do that) and he's not using homebrew.

pinepain:
@javian ptal. php-uv was recently tagged, so I added formulas for both 70 and 71 PHP.

Currently build fails because  `GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)` which seems to be ok.

pinepain:
@javian ping. External issue is resolved: php-uv tagged revision which we now use in this pr. PTAL.

pinepain:
Any chance to give someone a look at it?

pinepain:
@javian ptal this pr, please.

PR #3426, opened 2016-07-30
Add php70-phalcon formula

envomer:

-----


javian:
I think you have to go past 2008 to find a 32bit osx machine so I'm not sure how likely it is that the 32 bit section would be triggered. We also don't test it for 32bit so we wouldn't know if it works or not - do you think we should keep it still ? I could see that it would be useful in something like linuxbrew but i'm not sure how that interprets the macos checks.

javian:
Thanks @envomer ! Can you please use the right format for submitting new formulas in the future ?

```
==> Patch closes issue #3426
Warning: Nonstandard bump subject: Add php70-phalcon formula
Warning: Subject should be: php70-phalcon 3.0.0 (new formula)
```

envomer:
@javian Oh sorry about that, thank you and will do :D

PR #3483, opened 2016-08-17
php71-ref (new formula)

pinepain:

-----

This PR adds formula for [php-ref](https://github.com/pinepain/php-ref) PHP extension that adds soft and weak references to PHP (PHP 7 only). 

Disclaimer: I am the author of this php extension.


pinepain:
The reason of failure is `GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)`.

javian:
You seem to have chosen the same name as https://github.com/digitalnature/php-ref but since that's not an extension I don't think it will be a problem.

javian:
Thanks for the submission @pinepain - its now been merged.

pinepain:
I didn't even supposed that they exists until now. Hope there will be no confusion and https://github.com/digitalnature/php-ref won't mind such naming.

Originally php-ref was named php-weak (which was a php 7 impl. of https://github.com/colder/php-weakref with minor tweaks, before they added php7 support), but after soft refs was added recently, I had to come with a better naming, so `ref` is for **Ref**erence (basically, root namespace in this  ext is `Ref`) like java.lang.ref, but for PHP.

PR #3497, opened 2016-08-18
Add formula for `php70-gearman`

stephenorr:

This uses the GitHub release of wcgallego/pecl-gearman, in its recently released 2.0.1 stable version, which seems to work fine with PHP 7 (please note, it doesn't work with older versions of PHP; only with 7.0 onwards).

PECL itself seems to be largely dead these days, so this is one of the easier ways for mere mortals to get these packages installed and running. I've tested this on my local dev system running El Capitan, and I'll also try it on Sierra when I get home.
-----


javian:
It won't pass until the forked repo status is removed from the one that is used now. I've posted a request to do that here https://github.com/wcgallego/pecl-gearman/issues/33

javian:
But I will merge it for now since the rest passes so its mainly an annoyance.

javian:
Thanks for the submission @stephenorr ! Glad we made it here in the end =)

stephenorr:
Me too @javian - and it was my pleasure to help :)

Issue #3502, opened 2016-08-18
Update to the latest php71, php70 and php56


javian:
A reminder to update to the latest php versions. @pinepain can you assist with the latest 7.1 beta ?

pinepain:
Sure, will PR after trying it locally.

pinepain:
@javian done under PR #3503. php70 7.0.9 and php56 5.6.24 are at their latest version already.

javian:
@pinepain they tagged new versions yesterday in git so they are coming

pinepain:
@javian oh, I see, same as new beta.3. I totally forgot looking into fresh tags =). So, I guess we'll skip beta.2 for now, right?

javian:
@pinepain yes lets go straight to beta 3.

pinepain:
@javian do you need any assistance with 70 and 5.x?

javian:
@pinepain yes please put them in too - it would save a little time =)

pinepain:
Done under PRs #3505 and #3506.

pinepain:
Only 5.6 left. Is there the same issue like with 7.0 and 7.1?

javian:
Yes, no issue really - just needed time to complete the pr merge and submit the core extensions but I'll do that now.

pinepain:
Thank you!

javian:
On its way now - just missing #3517 where I missed a whitespace. I guess I could just have merged it without rebuilding it again but too late now =<

javian:
I'm closing this now since we are only 2 extensions away from completing the update. Thanks for the support @pinepain !

PR #3503, opened 2016-08-18
php71 7.1.0-beta.3

pinepain:

-----
As per https://github.com/Homebrew/homebrew-php/issues/3502.

This build is known to fail audit because of installing php headers.


pinepain:
brew bot build fails audit because of headers installed, but 

```
        ==> Fetching php71-intl from homebrew/php
==> Downloading https://homebrew.bintray.com/bottles-php/php71-intl-7.1.0-beta.2.el_capitan.bottle.tar.gz
==> Retrying download
==> Downloading https://homebrew.bintray.com/bottles-php/php71-intl-7.1.0-beta.2.el_capitan.bottle.tar.gz
Error: Failed to download resource "php71-intl"
Download failed: https://homebrew.bintray.com/bottles-php/php71-intl-7.1.0-beta.2.el_capitan.bottle.tar.gz
```

looks a bit suspicious for me - when we submit new version there can't be any bottles for it. Is the output above "false-positive"?

javian:
Can you bump the revision in 7.1.rb too ?

javian:
This seems to be how https://github.com/Homebrew/homebrew-php/commit/6591deddb3e52ef75fc4ce34b3dcb6e730c81fcc @ablyler has done it before.

pinepain:
Do we have to remove formula-level revision? I don't see that it was in use before and it is a bit unclear why it used, can you clarify this moment, please?

javian:
I was probably used to update the bottle at the time. There are 2 wyas of managing the botlte, remove the bottle code inside the bottle block or increase the revision. I think the instructions says that the bottle code block should be emptied but I might lean towards bumping the revision since that can be used for other things as well like a general bottle refresh.

pinepain:
Updated to beta.3 and emptied bottle code block.

It says that formula revision `Used for creating new Homebrew versions of software without new upstream versions.`, but I'm just not sure how "clean" to bump formula revision when formula is not actually changed. I see that when I removed formula revision audit said that you can't decrease formula revision (default revision is 0). But once again, I'm not as much into homebrew internals as I want, so if you will find some good link, maybe to some previous discussion, I would appreciate that.

javian:
But the upstream version has changed but the formula itself hasn't since we use variables for the version tracking. Therefore we need to bump the revision (or remove the code clock inside bottle but I would prefer revision for now) to force the rebuild of the bottles for the new version.

javian:
I see you bumped all formulas. I haven't thought of that scenario but do you think that that is necessary or do we only need to rebuild the extensions that are included in core ?

pinepain:
Yeah, makes sense, thanks for clarifying.

As to bumping all formulas, it's my bad actually. From what I know, it's not necessary to rebuild third-party extensions unless ABI was changed. I'll take bottle block back for non-core formula. Thanks for catching this.

javian:
This seems to timeout after 60 minutes because the build takes too long - we might need to split the core build and then do the extension updates in a seperate. MIght be other clues in the build log that i didnt have time to check now.

pinepain:
Ok, will split extensions out from this PR. Btw, is it ok to reset formula revision number when we increment version number?

pinepain:
@javian It looks like CI process designed to fail and we are fighting with it rather then using it as a servant. Here is what we have after extracting extension (gmp is used as an example):

```
==> brew fetch --retry php71-gmp
==> Fetching php71-gmp from homebrew/php
==> Downloading https://homebrew.bintray.com/bottles-php/php71-gmp-7.1.0-beta.3.el_capitan.bottle.tar.gz
==> Retrying download
==> Downloading https://homebrew.bintray.com/bottles-php/php71-gmp-7.1.0-beta.3.el_capitan.bottle.tar.gz
Error: Failed to download resource "php71-gmp"
Download failed: https://homebrew.bintray.com/bottles-php/php71-gmp-7.1.0-beta.3.el_capitan.bottle.tar.gz
==> FAILED
```

javian:
Yes. If the version number is bumped we can remove the revision.

About gmp I'm not sure what's going on to be honest. I guess it assumes the bottles are there since the sha is kept so it tries to download the "right" version according to the formula version and doesn't find any. That would mean we have to update the core extensions at the same time as was done in the example I pasted in earlier today from @ablyler .

javian:
Lets try adding the core formulas back but only for 1 of the new versions so that we don't overload the build bot. I can't recall if it runs multiple jobs at the same time but we might just as well try it.

pinepain:
`gmp` is not the only one formula that failed: basically, all formulas with `version PHP_VERSION` fails as that version increments, but no bottle available.

OK, so we add all ext formulas, but, say, only for php71, right?

pinepain:
Also, I would like to explicitly clarify about resetting formula revision number: when I did that, audit failed (see https://github.com/Homebrew/homebrew-php/pull/3503#issuecomment-240858489), but logically it looks like that it should be safe to reset revision when version got incremented.

javian:
As far as I know, removing the formula version when the software version is increased should not be a problem.

javian:
Looking at the build now `/usr/local/Cellar/php71/7.1.0-beta.3_2` seems to indicate that it thinks it is the second revision which doesn't match the formula in its current state. This seems a little weird to me.

javian:
Looks good so far but the time might still be an issue. Lets see how it goes.

pinepain:
It looks like it won't be able to pass.

`Build timed out (after 60 minutes). Marking the build as aborted.`

Are there any possibilities to increase this for specific build types? Probably, restrict such feature to limited number of users or so.

javian:
I think its tricky. I tried to look at the older commit and see how that had played out but the build log had been deleted so I could no longer see what happened. Looking at the commit log the bottles for php were generated at that initial commit and the extension bottles were put in afterwards.

javian:
I think I figured out why the bottle downloads fail when we rebuilt php71. HOMEBREW_DEVELOPER is set on the build system which means that if it tries to find a bottle and it fails it will fail hard. Usually on non-dev system it should build it from source if it can't find the bottle. So as long as the php71 bottle is built having the core extension bottles fail should be fine and then we can submit a revision bump for each of the core extensions to rebuild those bottles.

javian:
Thanks for trying and help sort this out @pinepain - closing this now since I've merged it through another PR.

PR #3509, opened 2016-08-20
PHP 7.1 Beta 3

javian:

PR #3510, opened 2016-08-20
php71 7.1.0-beta.3

pinepain:

-----

This PR is to experiment with splitting PHP core and extension formulas defs to be able to upgrade PHP version without hitting build timeout.


pinepain:
Build fails because of `Header files that shadow system header files were installed ...`, apart of that everything looks fine.

javian:
It might be a way to go just wondering how it could be managed effectively. If the goal with the ci is to prevent problems this only hides any problems that could pop up when upgrading.

pinepain:
Yeah, your way in #3509 is far better and reliable.

PR #3517, opened 2016-08-22
PHP 5.6.25 core extension bottle update

javian:


javian:
Having problems with the build bot again.

ablyler:
@javian no need to create revisions, as the version will change due to it pulling it in from the abstract.  also, i'll create separate PRs for each formula right now to get around the build bot issue.

javian:
@ablyler I think revisions make it easier to manage though. If a new version is released just update the revision and off we go. At least it makes more sense than adding a space to the end of the file imo =)

javian:
Since this has been merged through other PRs I will close this.

PR #3533, opened 2016-08-24
php56-intl: update bottle

ablyler:


ilovezfs:
If you aren't bumping the revision, you should use https://bot.brew.sh/job/Homebrew%20Testing/ to create bottles that are missing or need to be rebuilt.

PR #3542, opened 2016-08-25
remove rebuild from bottles

ablyler:


lferro9000:
Just to note that this is now breaking brew tap homebrew/php due to undefined method 'rebuild'.

https://github.com/Homebrew/homebrew-php/issues/3545

PR #3568, opened 2016-09-05
Use correct branch for PHP 7.1

ocean90:
When using `brew install php71 --HEAD` brew currently installs an old version because the [PHP-7.1.0 branch](https://github.com/php/php-src/tree/PHP-7.1.0) isn't updated since alpha3. The correct branch is [PHP-7.1](https://github.com/php/php-src/tree/PHP-7.1).


javian:
Thanks for pointing that out @ocean90 ! Merged.

PR #3574, opened 2016-09-15
Prevent php from hardcoding superenv SED path

javian:
There have been several issues related to this due to homebrew switching the directory structure without migrating old installs. This will eventually happen but it would be better practice to not be reliant on the superenv shim and work directly towards the bundled sed.

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -176,6 +176,9 @@ INFO
   end
 
   def install_args
+    # prevent php from hardcoding sed path from superenv
+    ENV["SED"] = "sed" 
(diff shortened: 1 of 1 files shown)


josegonzalez:
Shouldn't we just use system sed - `/usr/bin/sed` - instead of this?

javian:
This was suggest in the Slack chat and also used in other Formulas like https://github.com/Homebrew/homebrew-core/blob/master/Formula/libtool.rb

javian:
Looks like this doesn't work after testing it some more and the superenv is still chosen over the override.

PR #3604, opened 2016-09-23
Update php71 rc2 core extension bottles

javian:


javian:
Moved to individual submission due to timeout in brewbot

Issue #3686, opened 2016-10-07
Remove no longer supported PHP versions


alcaeus:
Note: this was previously discussed in #1327. Back then it was about bringing back PHP 5.2 and/or dropping PHP 5.3.

We have now reached the point where PHP 5.3, 5.4 and 5.5 have been EOLed, leaving 5.6 and 7.0 as the only supported PHP releases (with PHP 7.1 coming soon). Given that 5.3 has been EOLed more than 2 years ago and 5.4 has been EOLed more than 1 year ago, maybe it's time to start removing old formula?

One previous idea was to [drop support for 5.3 when 5.4 is EOLed](https://github.com/Homebrew/homebrew-php/issues/1327#issuecomment-56086692), which essentially means supporting one EOLed version and dropping everything that's older - in this case this would mean dropping 5.3 and 5.4, while leaving 5.5, 5.6 and 7.0 for now. What do you think?

javian:
We were discussing this briefly a little while back and @josegonzalez was thinking along the same lines. I would be fine to do it since it feels odd to support a no longer supported version. We were looking at the statistics for Wordpress - https://wordpress.org/about/stats/ - that shows that there is still a big portion of installations using non supported versions which continues to freak me out =)

One option we discussed was to move over the old formulas to a boneyard repo to keep them alive but not supported.

javian:
I'll see if I can compile some download statistics over the weekend.

alcaeus:
> One option we discussed was to move over the old formulas to a boneyard repo to keep them alive but not supported.

That's a good idea, that way people who have to support older PHP versions can still get packages, while we remove the overhead of updating formulae for dead PHP versions. I won't comment on Wordpress usage statistics - those freak me out too much to think about.

pinepain:
I would even move 5.5 to boneyard. If people want to use outdated version, they have to support them themselves. Vendor lock-in is not an excuse but a faulty policy of such vendors.

pinepain:
Based on stats, php53 and php56 could be removed without any doubts. php55 is also having negative dynamics and could be also considered as a candidate for removal from mainstream. It's nice to see that php71 is raising.

[php53 statistics:](https://bintray.com/homebrew/bottles-php/php53/view#statistics)
![php53](https://cloud.githubusercontent.com/assets/2185793/19200416/8edc3e10-8cd1-11e6-9d73-ccaf57468c48.png)

[...]

PR #3689, opened 2016-10-09
Remove PHP 5.3 and 5.4

alcaeus:
Discussion can be found in #3686.

Note: I have replaced `php54` in the installation instructions with `php70` but haven't tested it yet.


josegonzalez:
@javian can you figure out where we need to "boneyard" these to?

alcaeus:
I took a look at the [homebrew-boneyard](https://github.com/homebrew/homebrew-boneyard) repository, we'd completely swamp it with PHP stuff. How about creating a new homebrew-php-boneyard or homebrew-php-unsupported repository and move everything that is no longer supported there?

javian:
@alcaeus I agree that the regular boneyard probably wouldn't be a good option.
My current idea was to add a caveat text about it no longer being supported and separate the core code into a boneyard version of the abstract files so that it could easily be moved at a later stage whilst us being able to ignore older versions in the abstract code. That would also prepare for future bone yarding so we don't get conflicts with abstract-php for instance when transferring the next php version into the boneyard. Alternatively we could rename the abstract file(s) when the bone yarding takes place since the current snapshot ought to work.

alcaeus:
\*bump\*

if someone could create the repository I'd be glad to do the rest of the work of moving the formulae to that other repository so we can remove them from homebrew-php.

alcaeus:
@javian What are we going to do about these? I'm still in favor of removing them once and for all.

alcaeus:
Closing since the problem will solve itself with https://github.com/Homebrew/homebrew-core/pull/16067. It's not worth the rebases.

PR #3440, opened 2016-07-31
Add php71-mustache formula

pinepain:
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-php/pulls) for the same formula update/change?
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

-----

PR #3723, opened 2016-10-23
libmustache 0.4.2

rwhogg:

-----

Use Git for the download so the submodules are included. (Without them, autoreconf fails).


rwhogg:
Question: should I also bump the php{54,55,56,70}-mustache formulae to version 0.7.2? The versioning scheme suggests there shouldn't be any API breaks, and I can confirm that the existing php56-mustache formula still compiles & test passes with v0.4.2, but it's little to no additional work to add those. 

(Note that php71-mustache was already bumped in #3440).

javian:
Yeah it looks like that the the related formulas should be revisioned. Can you submit that ?

javian:
@BrewTestBot test this please

javian:
Thanks for the submission @rwhogg !

rwhogg:
> Yeah it looks like that the the related formulas should be revisioned. Can you submit that ?

Will try to get to that this weekend 😄

rwhogg:
> Yeah it looks like that the the related formulas should be revisioned. Can you submit that ?

See #3745

Issue #3590, opened 2016-09-20
Brew provides shell script as Composer executable, other tools expect a PHP script


bocharsky-bw:
## Description

Brew provides broken `composer.phar` file. It works in regular use, but does not work with Ansible. If I replace `/usr/local/Cellar/composer/1.2.1/libexec/composer` (which provided by Brew) with the original `composer.phar` downloaded with [these instuctions](https://getcomposer.org/download/) - then original Composer works fine with Ansible. So the problem only in `composer.phar` provided by Brew.
## Steps to reproduce:
1. `$ brew install ansible`
2. `$ brew install composer`
3. `$ ansible localhost -m composer -a "working_dir=./path-to-dir-with-valid-composer-json"` // it doesn't work, see logs below:
## Context and Logs

Output of `$ ansible localhost -m composer -a "working_dir=./path-to-dir-with-valid-composer-json"`:

```
Loaded callback minimal of type stdout, v2.0
<127.0.0.1> ESTABLISH LOCAL CONNECTION FOR USER: victor
<127.0.0.1> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo $HOME/.ansible/tmp/ansible-tmp-1474021242.83-74316657830341 `" && echo ansible-tmp-1474021242.83-74316657830341="` echo $HOME/.ansible/tmp/ansible-tmp-1474021242.83-74316657830341 `" ) && sleep 0'
<127.0.0.1> PUT /var/folders/5t/qb0jlm4x7g90bp__cnr1lybh0000gn/T/tmpLAYqVM TO /Users/victor/.ansible/tmp/ansible-tmp-1474021242.83-74316657830341/composer
<127.0.0.1> EXEC /bin/sh -c 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python /Users/victor/.ansible/tmp/ansible-tmp-1474021242.83-74316657830341/composer; rm -rf "/Users/victor/.ansible/tmp/ansible-tmp-1474021242.83-74316657830341/" > /dev/null 2>&1 && sleep 0'
An exception occurred during task execution. The full traceback is:
Traceback (most recent call last):
  File "/var/folders/5t/qb0jlm4x7g90bp__cnr1lybh0000gn/T/ansible_a_ligN/ansible_module_composer.py", line 232, in <module>
    main()
[...]

PR #3724, opened 2016-10-23
composer: use Phar::loadPhar() instead of passthru()

benjifisher:

---

### DIFF
--- a/Formula/composer.rb
+++ b/Formula/composer.rb
@@ -6,8 +6,8 @@ class Composer < AbstractPhpPhar
   homepage "http://getcomposer.org"
   url "https://getcomposer.org/download/1.2.1/composer.phar"
   sha256 "c2e04040b807a8530e5c83de56bdaaf63a0f183f8fd449bbe6e41f660e647427"
-  head "https://getcomposer.org/composer.phar"
   revision 1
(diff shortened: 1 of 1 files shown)


benjifisher:
This gives a better fix for #3590.

There is also an unrelated change in order to make the change pass `brew audit --strict --online`.

javian:
This seems ok to me so I would suggest we merge this and consider to create a more general solution in the #3725

benjifisher:
I am  having second thoughts about this solution.  Looking more closely at `composer.phar`, I see that it has some lines that test APC version and configuration, and then the lines

```php
Phar::mapPhar('composer.phar');
require 'phar://composer.phar/bin/composer';

__HALT_COMPILER(); ?>
```

followed by the binary part of the archive.  I want to understand the difference between `mapPhar()` and `loadPhar()`, and I want to make sure we do not lose the checks related to APC.

javian:
@benjifisher have you gotten any further with this ?

javian:
No feedback on this so I'm going to close it for now - feel free to reopen if you have any further ideas.

javian:
According to the php docs on mapPhar `mapPhar should be used only inside a phar's loader stub. Use loadPhar to load an external phar into memory.` so I would not use that.

PR #3735, opened 2016-10-25
wp-cli 0.25.0

danielbachhuber:


javian:
Thanks for the submission @danielbachhuber !

PR #3749, opened 2016-10-31
Adds club formula

kylebrowning:

-----

This adds club the command line utility for BLT, a build and launch tool for Drupal projects.


josegonzalez:
@javian I'm 👎 on adding more phar formula for installation. Especially one with limited use cases - in this case it seems it's only for furthering the Drupal community.

kylebrowning:
There are already three Drupal related utilities in here. 

https://github.com/Homebrew/homebrew-php/blob/master/Formula/drupal-code-sniffer.rb
https://github.com/Homebrew/homebrew-php/blob/master/Formula/drupalconsole.rb
https://github.com/Homebrew/homebrew-php/blob/master/Formula/drush.rb

josegonzalez:
@kylebrowning I'd also be in favor of removing those as well - and anything that is framework-specific, or not of use to the community at large.

That said, @javian and @ablyler are the current active maintainers, so I'm only voicing my opinion as the repo creator. If they decide otherwise, thats where the project should go.

chrowe:
For me I find it very helpful to be able to install these tools with brew.

And there are framework specific tools already in here like https://github.com/homebrew/homebrew-php/blob/master/Formula/wp-cli.rb and https://github.com/Homebrew/homebrew-php/blob/91c73afe66b3b605730fc423a0e503002cf002f5/Formula/symfony-installer.rb

josegonzalez:
@chrowe I'm also suggesting we remove those as well. If they were moved to a `homebrew-php-phars` repo, that would be fine.

My reasoning is that it's not all that hard to install a phar, and phars move much more quickly than anything else, so are a maintenance burden. Creating a standalone phar tap would also allow us to include more phars from more communities, while adding people who want to be phar-tap maintainers. The workflow is a bit different from the core of homebrew-php.

Again, up to @javian and @ablyler, I'm only making my opinion heard :)

grasmash:
Just to throw my hat in the ring-- I find it tremendously useful to install framework-specific tools via homebrew. Whether they are part of homebrew-php tap or a hombebrew-php-phars has no impact on my use case.

kylebrowning:
Also of note are the Abstract classes and requirements for Phars which is maintained here. Would we need to replicate them, or can they be shared?

javian:
@kylebrowning it would work a standalone tap to require the files from the php tap in the same way that phars are constructed but the only way to do that as far as I know would be to require one of the Formulas in here which would trigger the tap which might not be ideal.

kylebrowning:
@javian Can you elaborate or provide a short example? are you referring to depends_on?

javian:
As far as merging this I would be against it for now. Brew bot throws out these messages
```
homebrew/php/club:
  * GitHub repository not notable enough (<20 forks, <20 watchers and <50 stars)
  * GitHub repository too new (<30 days old)
  * fully scope test system calls e.g. system "#{bin}/club"
```
The reasoning for this would go along these lines that the tool is too new and don't have enough traction in the community whereas something like drush is widely established in the community already. I would recommend setting up your own homebrew tap and put your formula in there for now.

javian:
@kylebrowning sure! The line in the formula that does all the work are
```
require File.expand_path("../../Abstract/abstract-php-phar", __FILE__)
```
You can adapt this to a tap that you create yourself by adjusting the path but there's no way that I know of to create a dependency to a tap - only to a Formula inside a tap (through `depends_on`). So, if you in your instructions say that the php tap needs to be tapped then it probably shouldn't be an issue.

kylebrowning:
Okay happy to close this PR.

kylebrowning:
```
require File.expand_path("/usr/local/Homebrew/Library/Taps/homebrew/homebrew-php/Abstract/abstract-php-phar", __FILE__)
require File.expand_path("/usr/local/Homebrew/Library/Taps/homebrew/homebrew-php/Requirements/composer-requirement.rb", __FILE__)
```

Side question tho, @javian Anyway to get home-brew tap path incase its not /usr/local/Homebrew?

josegonzalez:
`brew --prefix` will give you `/usr/local` as the prefix install path.

kylebrowning:
I meant `#{HOMEBREW_REPOSITORY}` for inside the formula.

javian:
You could use the same method as we do with the formulas using a relative path. The tap directory won't change location so if you make it relative then you should  be fine.

chrowe:
@kylebrowning are you planning to create your own tap for this? if so, can you provide a link here when it is ready.

kylebrowning:
@chrowe Yes, its coming very soon, just needing to get Acquia approval.

PR #3770, opened 2016-11-13
php71 rc6 core extension revision bumps

javian:


ilovezfs:
@BrewTestBot test this please

PR #3771, opened 2016-11-15
New Formula Deployer 4.0.0

anthonysomerset:
Yes - there are none
Yes - works fine

```10:58 $ brew install deployer
==> Installing deployer from homebrew/php
==> Downloading https://deployer.org/releases/v4.0.0/deployer.phar
######################################################################## 100.0%
🍺  /usr/local/Cellar/deployer/4.0.0: 5 files, 3.1M, built in 50 seconds
```
Yes - No Errors
-----

Deployer PHAR v4.0.0, may possibly need the same hack as composer referenced in https://github.com/Homebrew/homebrew-php/issues/3590

Otherwise works great on local CLI


josegonzalez:
@MikeMcQuaid if we wanted, could we create a homebrew-phars repo and have it be official?

MikeMcQuaid:
@josegonzalez What would be the advantage over having it just in e.g. homebrew-php?

javian:
@MikeMcQuaid José is giving some of his arguments in here https://github.com/Homebrew/homebrew-php/pull/3749 . I believe it doesn't give any advantage other than having a different set of maintainers working on the phars exclusively. I think we should be packaging the most popular tools despite them being accessible and able to be installed through other means because I'm assuming that is what the community would ask for.

MikeMcQuaid:
> I believe it doesn't give any advantage other than having a different set of maintainers working on the phars exclusively.

We've had pretty bad experiences with this in homebrew/core with e.g. homebrew/versions etc. I think it's worth consolidating communities on single taps around groups of maintainers. We'll be pulling homebrew/versions, homebrew/x11, homebrew/gui etc. back into homebrew/core eventually as a result.

josegonzalez:
I'm just of the opinion that these phar bits move too quickly for us to be packaging here - though if someone wants to maintain that (like the OP of this PR) then thats okay with me.

It's kinda like create a homebrew formula for a ruby gem or npm package.

MikeMcQuaid:
> It's kinda like create a homebrew formula for a ruby gem or npm package.

We have a bunch of them in homebrew/core 😁. My take these is: if they are useful as an application (not library) and people don't really care what package management system they are from: we ship them. In a tap like this my take would be it's reasonable to add applications and libraries that are depended on by other formulae but nothing else.

josegonzalez:
Sad panda. Hokay, you've convinced me, and I won't bring up the topic again.

anthonysomerset:
i don't think theres enough critical mass for seperate phar tap here - so this is the next best logical place for them

my only logic behind this phar in particular is specifically that install instructions say to copy to /usr/local/bin which is where my homebrew stuff is - i'd much rather avoid putting extra stuff other than homebrew in here if i can help it - this phar doesn't seem to be updated too often to become too significant a maintenance burden (at this stage)

javian:
@anthonysomerset thanks for your contribution! I saw you pointed out in the commit that it might need something similar to the composer "hack" - can you verify that ?

PR #3775, opened 2016-11-15
php-56 5.6.29

javian:


kyleoliveira:
Any idea why this failed? It would be really awesome to get this in, but I don't see what's wrong.

javian:
The core formulas always fails since they include the php headers which conflicts with what's bundled in osx. I'll see if we can get this merged tomorrow.

javian:
@BrewTestBot test this please

Issue #3823, opened 2016-12-08
yaml 2.0 is out


acim:
## Description

So it would be nice if someone could update 7.x formulas (php70-yaml.rb and php71-yaml.rb). Thanks.

josegonzalez:
Pull requests welcome.

acim:
Will try to do it right now :)

javian:
Thanks @acim the PR has been merged now.

PR #3824, opened 2016-12-08
#3823 Updated PHP 7.1 yaml formulas to use PECL yaml 2.0.0

acim:


javian:
Thanks for you submission! You do not have to touch any of the bottle code nor add it if it is missing. The brew bot will take care of all that for us.

PR #3827, opened 2016-12-09
php70 7.0.14

alcaeus:

-----


javian:
Thanks @alcaeus ! I'll get this merged asap - also releasing 5.6.29 today so the queue is a bit longer than usual =)

alcaeus:
@javian no worries, thanks! 🍻

javian:
We need to address all these brewbot problems some day to make it easier to determine if we have any issues with the build =)

alcaeus:
@javian yes, so I've noticed. BTW, I just wanted to push the revision updates for the core extensions since I forgot those in the first go. Do you want me to create a new PR or do you want to fix them directly?

javian:
@alcaeus Thanks for offering. I was thinking that we perhaps should include the core extensions in the commit when we upgrade the version to include everything in one go. I think I have experimented with that in the past and its very much up to the load on the brewbot if it works or not since it extends the time it takes to produce the bottles but I think I would prefer that to make it a little bit more easy to manage. What do you think ?

alcaeus:
I agree, it should be in one commit. The build architecture should be made to deal with it. After all, updating the core extensions is part of updating the php formula, without it the commit is technically broken.

PR #3834, opened 2016-12-09
Fix issue with dependency resolution when used by a head-only formula

mgoeppner:
# Bug Fix

This fixes an issue where `httpd24` (or `httpd22`) and it's dependencies would not be installed correctly if `php54` is used as a dependency of another formula in such a manner:

```
depends_on 'php54' => ['with-apache']
```

The patch deprecates `--with-apache` and `--with-apache22` and replaces them with `--with-httpd24` and `--with-httpd22`.

> The problem is that Homebrew options that aren't bound to a formula aren't really first-class citizens.

See https://github.com/Homebrew/homebrew-core/issues/7719 for additional details.  @ilovezfs was kind enough to put together a patch and the quoted explanation.

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -42,9 +42,6 @@ class AbstractPhp < Formula
     depends_on 'unixodbc' unless build.include?('without-unixodbc')
     depends_on 'readline'
 
-    depends_on 'homebrew/apache/httpd24' if build.include?('with-apache')
-    depends_on 'homebrew/apache/httpd22' if build.include?('with-apache22')
-
     # ssl
     if build.include?('with-homebrew-libressl')
       depends_on 'libressl' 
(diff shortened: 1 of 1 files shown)


javian:
Thanks for putting this together @mgoeppner - I learned something new tonight =)
I also see that there are a few other options built up in the same way so that should probably be reviewed and adapted as well at one point.

mgoeppner:
@javian I made an issue to track that work -- if you know of any places in particular that need that change, could you please add them to the issue?

Cheers!

javian:
Absolutely!

Abstract/abstract-php.rb
```diff
@@ -61,10 +58,15 @@ def self.init
...
+    option 'with-httpd24', 'Enable building of shared Apache 2.4 Handler module'
+    option 'with-httpd22', 'Enable building of shared Apache 2.2 Handler module'
     deprecated_option "homebrew-apxs" => "with-homebrew-apxs"
-    deprecated_option "with-homebrew-apxs" => "with-apache"
+    deprecated_option "with-homebrew-apxs" => "with-httpd24"
+    deprecated_option "with-apache" => "with-httpd24"
+    deprecated_option "with-apache22" => "with-httpd22"
+
+    depends_on "homebrew/apache/httpd24" => :optional
+    depends_on "homebrew/apache/httpd22" => :optional
```
axiac: These two lines make the formula always depend on the `homebrew/apache` tap, even when nothing is required from it.
This change rendered installing or updating PHP impossible on Linuxbrew because `homebrew/apache` cannot be installed on Linuxbrew.

Shouldn't these two lines be like this?

    depends_on "homebrew/apache/httpd24" => :optional if build.include?('with-httpd24')
    depends_on "homebrew/apache/httpd22" => :optional if build.include?('with-httpd22')

I cannot tell anything about the situation when PHP is a dependency of other formula but the change I propose here fixes the installation and upgrade of PHP on Linuxbrew when the `with-httpd2x` options are not used.

axiac: On a second thought, Homebrew works fine after this change, it can install or upgrade PHP without installing the `homebrew/apache` tap (when none of the `with-httpd2x` option is used). I think this means that Linuxbrew is faulty now and it should catch up with the recent changes in Homebrew core.

rwhogg: > I think this means that Linuxbrew is faulty now and it should catch up with the recent changes in Homebrew core.

That's possible; we're currently 2 stable releases behind Homebrew. @linuxbrew/core does someone want to release at least 1.1.7 and ideally 1.1.8 this weekend?

PR #3836, opened 2016-12-11
composer 1.3.0-RC

alcaeus:

-----


javian:
The devel version doesn't seem to be a phar and needs to be compiled to work as expected.

alcaeus:
Oh, right, I didn't see that. I'll get on it tomorrow morning.

javian:
This versioning scheme might cause problems later but we'll see when we get there.

javian:
Thanks @alcaeus !

alcaeus:
> This versioning scheme might cause problems later but we'll see when we get there.

You mean just calling it "RC" instead of "RC-1"? I'd think if they need another release candidate, that would be "RC-2" to avoid confusion.

javian:
The problem is that brew discards anything after the 0. Looking at their commit history they only released one RC for 1.2.0 so it should probably be ok

PR #3860, opened 2016-12-19
php70-memcache (new formula)

GedRap:

-----

The formula does not pass strict audit because there is no stable tag, similar to existing `php-memcache`therefore I assumed that this is acceptable, although not great.

The only potential issue is that it uses not the original source code from PECL but a forked version since the code in PECL does not support php7 and it seems like it is not in the plans to do so (the beta version is in beta for the past 3 years).

However, seems like there are no other options because plenty of code in the wild depends on memcache.

Please let me know if I could improve the PR.

Thanks.

### DIFF
--- a/Formula/php70-memcache.rb
+++ b/Formula/php70-memcache.rb
@@ -1,0 +1,21 @@
+require File.expand_path("../../Abstract/abstract-php-extension", __FILE__)
+
+class Php70Memcache < AbstractPhp70Extension
+  init
+  desc "Interact with memcached through OO and procedural interfaces."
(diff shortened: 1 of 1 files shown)


GedRap:
Could some please provide me some feedback how to improve this PR?

javian:
Hi @GedRap - thanks for you submission. There are a few issues with this PR as I see it. Its not based on an official release and it uses the same name as been used previously and it also has no released versions. Unless these 3 parameters change then I would not want to include it here.

glensc:
it's defacto repo for php7 for this module

javian:
Closing since the formula has now been merged #4025.

PR #3893, opened 2017-01-04
Move include directory to a custom path

javian:
This will make php builds pass the audit check when upgrading
or bumping revisions.

-----
This is an attempt to bypass the audit check in homebrew to able to properly test php builds since they all currently fail by default when the audit checks for the include files. This might not be merged after talking to @MikeMcQuaid who suggested that whitelisting the formulas might be another possible option since it could affect other modules and software.

As far as I can tell it properly updates the php-info and phpize commands which are both used to create the environment for compiling php modules. There might be other scenarios that fall outside of this. It might also lead to that we need to recompile all modules in the tap which also does not seem feasible.

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -366,7 +366,11 @@ INFO
 
   def _install
     system "./buildconf", "--force" if build.head?
-    system "./configure", *install_args()
+    # Pass custom includedir option to pass homebrew audit check
+    system "./configure", *install_args(), "--includedir=" + prefix/"homebrew_include"
+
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> This is an attempt to bypass the audit check in homebrew to able to properly test php builds since they all currently fail by default when the audit checks for the include files. This might not be merged after talking to @MikeMcQuaid who suggested that whitelisting the formulas might be another possible option since it could affect other modules and software.

Yeh, I'd recommend we whitelist these in `brew audit` instead if they aren't causing any user-reported problems.

PR #3905, opened 2017-01-10
php70-opcache: new testing method for module

javian:
Testing a new test method to see if bot picks up on compiled module issues.


javian:
This will no longer be necessary to implement since we are moving the Formulas to core.

PR #3906, opened 2017-01-10
terminus 1.0.0-beta.2

BePsvPT:
---


javian:
I don't think we should replace this since it breaks backwards compatibility. We could add the beta as a "devel" option in the formula and once 1.0 is released see if we need to split the formulas to keep the old version around for a while.

javian:
Thanks @BePsvPT !

PR #3936, opened 2017-01-23
phpmyadmin RELEASE_4_6_6

BePsvPT:
---


javian:
@BePsvPT I think your tool got the version for phpmyadmin incorrectly and it should match the other formulas ^^ I'll fix it now when merging - thanks again for you submissions.

PR #3937, opened 2017-01-24
terminus@0.13 0.13.6 (new formula)

jribe:

-----

Terminus 1.0 is a major breaking change and some people/organizations (like me) need access to 0.x while we upgrade.

### DIFF
--- a/Formula/terminus@0.13.rb
+++ b/Formula/terminus@0.13.rb
@@ -1,0 +1,36 @@
+# This formula tracks the 0.13 branch of Terminus
+# Terminus 1.0 is a major breaking change and users may need time to adapt
+require File.expand_path("../../language/php", __FILE__)
+require File.expand_path("../../Requirements/php-meta-requirement", __FILE__)
+
+class TerminusAT013 < Formula
+  include Language::PHP::Composer
+
+  desc "Command-line interface for the Pantheon Platform"
+  homepage "https://github.com/pantheon-systems/terminus"
+  url "https://github.com/pantheon-systems/terminus/archive/0.13.6.tar.gz"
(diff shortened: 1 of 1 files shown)


jribe:
My cursory scan of the logs looks like the Jenkins failure may have been an intermittent issue. Can somebody rekick it?

javian:
@jribe I'll trigger the build once more but you right - it seems to have been a CI issue not related to the code. I would however like to ask you to rename the formula to  `terminus@0.13` to keep it more in line with other versioned formulas.

jribe:
Renamed to `terminus@0.13`

javian:
Thanks for the submission @jribe !

javian:
@jribe I just realized that this now conflicts with the `terminus` formula since they both share the same binary name. Would you prefer that a `conflicts_with` is added to prevent this or should we rename the binary for the legacy terminus when its installed ?

javian:
And just for claritys sake, if conflicts_with is added then you can't have both formulas installed at the same time-

jribe:
`conflicts_with` would be preferable for our usage. I would've added that myself, but I wasn't sure how.

javian:
@jribe here's an example
`conflicts_with "php54-apcu", :because => "Provides conflicting functionality"` from https://github.com/Homebrew/homebrew-php/blob/master/Formula/php54-apc.rb . Here's another one from core `conflicts_with "jam", :because => "both install a 'jam' binary"` from ftjam.rb

javian:
@jribe #4022 is the suggested way of doing it. This will produce an additional line in the caveats saying
```
This is an alternate version of another formula.

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/terminus@0.13/bin:$PATH"' >> ~/.zshrc
```
and resolve the conflict which I think would be sufficient.

PR #3977, opened 2017-02-08
Use pecl package for imagick formula

alcaeus:

-----

This fixes an issue introduced in #3970. Using the github archive works, but the module reports its version as `@PACKAGE_VERSION@`:

```
$ php -r "var_dump(phpversion('imagick'));"
Command line code:1:
string(17) "@PACKAGE_VERSION@"
```

This causes issues when relying on a specific version, e.g. when having a composer version constraint for `ext-imagick`. Using the packages from PECL fixes this error.


javian:
Alternatively we could patch https://github.com/mkoppanen/imagick/blob/6cda6150bd98673b00310a5e38274bf1bd1a84d3/php_imagick.h since using head would still continue to function in the same way you describe but I'm not sure how to properly get the version from head.

javian:
The version number is stored here https://github.com/mkoppanen/imagick/blob/6cda6150bd98673b00310a5e38274bf1bd1a84d3/package.xml

alcaeus:
Patching it would mean having to update the version number on every update, with little to no advantage compared to downloading the package from pecl.

I'm not sure what we'd use for package_version when installing head.

javian:
We wouldn't have to update the version number since its derived from the URL and it can be accessed by using the right variable assuming it matches the release version.

javian:
head would be a bit more tricky since we would have to read the package.xml file and get the version from there.

javian:
I can merge this now to fix the issue and we can consider doing this later if necessary.

alcaeus:
Sounds good - I'll take a look at patching the package version.

javian:
We just need one more change. Can you please add a revision number for each formula ? If not then it wouldn't trigger a re-download for the people who already have it installed.

alcaeus:
Done.

javian:
Thanks @alcaeus !

PR #3996, opened 2017-02-17
README: Remove versions tap, fix spelling+wording

neutric:
Remove deprecated homebrew/versions tap and fix spelling and wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,8 @@ Please refer to [this section](#filing-bug-reports) for more information.
 
 Bugs inevitably happen - none of us are running EVERY conceivable setup - but hopefully the install process can be made smoother through the following tips:
 
-- If you have recently upgraded your Mac OS version or Xcode, read [this section](#common-upgrade-issues).
-- Upgrade your OS X to the latest patch version. So if you are on `10.9.0`, upgrade to `10.9.2` etc.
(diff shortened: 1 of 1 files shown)


javian:
Thanks for this @neutric . As far as I know you are not required to tap the dupes tap anymore but that might have changed recently - I haven't tested it in a while. Can you confirm that that is the case ?

neutric:
I remember having read something somewhere, but wasn't able to find it just now, so I tried it out by installing another instance of Homebrew to a non-default location (~/homebrew) and then installing something which requires a formula in `homebrew/dupes`.

```bash
% ~/homebrew/bin/brew install htop --with-ncurses
==> Tapping homebrew/dupes
Cloning into '/Users/ah/homebrew/Library/Taps/homebrew/homebrew-dupes'...
```

As you can see, `homebrew/dupes` is tapped automatically. I can't rule out that there is a variable set by my main homebrew installation which influences which/if taps are autotapped.

@ilovezfs probably knows.

javian:
@neutric can you sort out this conflict ? There's been a few changes to the readme file in other commits.

neutric:
I will soon, just short on time atm.

neutric:
@javian Finally fixed. Sorry for the loooooong delay!

javian:
Thanks @neutric ! Better late than never =)

PR #4001, opened 2017-02-19
Update ISSUE_TEMPLATE

neutric:


javian:
I think we should replace whatever we can in this instruction with the output of `brew config` instead which replaces most of this info. Can you make the update ?

javian:
Thanks @neutric !

PR #4012, opened 2017-02-26
Updating Phalcon

richmilns:
5.3 => 2.0.13
5.5 => 3.0.4
5.6 => 3.0.4

- [ Y ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-php/pulls) for the same formula update/change?
- [ N ] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- [ ? ] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

-----


javian:
@SrihariThalla we'll deal with that when we boneyard those php versions, for now lets keep it. @richmilns would you know if 7.1 is also supported ?

Formula/php53-phalcon.rb
```diff
@@ -4,8 +4,8 @@ class Php53Phalcon < AbstractPhp53Extension
   init
   desc "A full-stack PHP framework"
   homepage "http://phalconphp.com/"
```
javian: Homepage needs to be https://

Formula/php55-phalcon.rb
```diff
@@ -4,8 +4,8 @@ class Php55Phalcon < AbstractPhp55Extension
   init
   desc "Full-stack PHP framework"
   homepage "http://phalconphp.com/"
```
javian: Homepage needs to be https://

Formula/php56-phalcon.rb
```diff
@@ -4,8 +4,8 @@ class Php56Phalcon < AbstractPhp56Extension
   init
   desc "Full-stack PHP framework"
   homepage "http://phalconphp.com/"
```
javian: Homepage needs to be https://

javian:
@richmilns now that I was looking at the php5.3 compile errors. Unless you can fix them (easily) then I would just exclude php53 from this commit. There's been a whooping 2 installs over the last month of the 5.3 version and if they care enough then they need to fix it themselves.

richmilns:
@javian Phalcon 3.1 I believe will support PHP 7.1. At the moment 3.0.x does not support 7.1. I will have a look into the errors.

richmilns:
I couldn't see the reason why 5.3 was failing to build, so I've rolled it back to the previous version and sha256 value. Thanks!

javian:
Thanks @richmilns !

PR #4019, opened 2017-03-01
Moved Requirements and Installation sections to top

Zifius:
Reinstalled homebrew-php on a new laptop and it is a bit disturbing that the Installation section doesn't come on top ;)


javian:
Thanks @Zifius !

Zifius:
My pleasure @javian!

PR #4025, opened 2017-03-06
Add php70-memcache

mtibben:
It seems the memcache extension has not been packaged here for PHP 7 because the 2.x.x version fails to compile. However, I noticed [ubuntu do have it packaged](https://launchpad.net/ubuntu/+source/php-memcache) and found they were using a forked repo https://github.com/websupport-sk/pecl-memcache. Fedora seem to be using it too.

So here I've taken the same approach and am installing memcache from the fork.

This looks similar https://github.com/Homebrew/homebrew-php/pull/3860

### DIFF
--- a/Formula/php70-memcache.rb
+++ b/Formula/php70-memcache.rb
@@ -1,0 +1,22 @@
+require File.expand_path("../../Abstract/abstract-php-extension", __FILE__)
+
+class Php70Memcache < AbstractPhp70Extension
+  init
+  desc "Memcache PHP extension"
+  homepage "https://github.com/websupport-sk/pecl-memcache"
+  url "https://github.com/websupport-sk/pecl-memcache/archive/4991c2fff22d00dc81014cc92d2da7077ef4bc86.tar.gz"
+  version "3.0.9.20160311"
+  sha256 "498ae2f3fde66b471eb8662efedfe5cb2f89f5fff725791ffddbcc7d94b7bf21"
+  head "https://github.com/websupport-sk/pecl-memcache.git", :branch => "NON_BLOCKING_IO_php7"
+
+  def install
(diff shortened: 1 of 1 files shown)


mtibben:
Ok I've updated this to use, [the same version](https://github.com/websupport-sk/pecl-memcache/tree/4991c2fff22d00dc81014cc92d2da7077ef4bc86) as [Ubuntu](https://launchpad.net/ubuntu/+source/php-memcache) and [Fedora](https://apps.fedoraproject.org/packages/php-pecl-memcache/).

This also fixes linting

javian:
@mtibben thanks for this. Due to the popularity of this module I think we should merge it. We should add a caveat section about this though since its not the official module that is used and perhaps add  some comments in the formula as to where we can track the versions used by the linux distributions.

One note about using version the way you did - anything after the dash `-` will be ignored so it won't pickup on that part. I would suggest that we adopt the format `version "3.0.9.20160311"` instead or the version line will be useless.

Can you please also make the commit message `php70-memcache 3.0.9.20160311 (new formula)" ?

mtibben:
Sure thing.

So to be clear `3.0.8` was actually the _last_ official release. 3.0.9 is completely made up by Fedora (`3.0.9-0.3.20160311git4991c2f.fc26`) and Ubuntu (`3.0.9~20160311.4991c2f-5`). I'm guessing it's done so if a `3.0.9` is released it still lexicographically sorts as being greater than `3.0.9~.....` So when comparing version numbers `3.0.9` comes out on top.

So given that, `3.0.9.20160311` should be fine too

mtibben:
Done

javian:
@mtibben thanks! Did you ever add a caveat ? Its not in the formula now.

mtibben:
No caveat. Thanks!

PR #4026, opened 2017-03-07
php-couchbase: update to 2.3.0

avsej:

-----


javian:
Please exclude removing the php53 formula in this commit.

avsej:
Why? We do not support php 5.3, successful build does not mean it work there properly

avsej:
@javian I did all the checks requires in the PR template. Also strictly speaking we don't support php 5.5, but I will assume it will at least work there as we support 5.4 and 5.6

javian:
My point is that I don't want to remove formulas in an upgrade commit - it should be handled separately. I'd also rather have a new version like the one you submitted for 7.1 in a separate commit to be able to pay attention to that specifically and not get in the way for upgrading formulas that already exist. You don't have to change the 7.1 one formula this time since I'm aware of it now.

avsej:
@javian I've updated it. I will remove 5.3 in next PR

javian:
@avsej you also don't have to remove the bottle lines in the PR, the bot will take care of all of that and it makes the commits smaller. thanks for the submission!

Issue #4005, opened 2017-02-20
The future of the PHP tap and formulas, please give your input


javian:
This has turned into a slightly longer text than I would have hoped but I would like to get some feedback. Using an issue might not be the best way to collect this kind of information but it is a start and I'd like to keep it as open as possible for people who input. I also don't know how many people in the community that this will touch but there are other ways of reaching out in case it is needed. The text is not super structured but I have been sitting on this for a while to try and get some input and at this stage it was more important to get it our rather than it being perfect so if something doesn't make sense that's probably why =)

I’ve been thinking about how we can move the PHP tap forward since I currently feel that the code is a bit messy and that the main core formulas really belong in the upstream brew core project and I’ve had a short discussion with Mike about it and he agrees that the “popular” packages should/could be moved across. As I see it that would mean PHP 5.6, 7.0 and 7.1 and any subsequent version released of PHP. Potentially it could be discussed if we should go further back in the PHP release timeline but as we’ve already discussed boneyarding the php 5.3 and 5.4 formulas due to them not being supported upstream then it is probably better to spend that time moving forward.

To be able to move the main php formulas into core we would have to remove the abstractions and make the PHP versions into standalone formulas as well as adopting the new version format used in brew ie.  **php@7.0** . I think this would not be a big issue since the formula code does not change very often and we would also gain the following benefits

_Advantages_
- Easier to read Formula code
- Easier to read if a formula has been updated
- Revisions are currently the only mechanism to upgrade a formula and it is not very clear whether an extension has been updated due to the version abstraction currently in place.
- More maintainers that can merge and maintain updates
- Possibility of creating a new `:php` dependency that would also support versions similar to other programming languages in brew. 
- Possibility of creating core code to support phars and extensions.

_Drawbacks_
- It would increase the time it would take to commit a new PHP version with the current structure but I think that the formula clarity would be preferred over saving a little bit of time. This could of course be changed as will be outlined below.
- Code duplication since a lot of code in the php formulas is shared

The other thing that would need to happen is to decide on how we would want the PHP versions/SAPIs to work in brew. Currently we are supporting the main formula phpXX.rb which includes several different SAPIs (cli, apache, fpm etc) and there has been a previous attempt to split these into separate files which I still think might be a good idea but introduces other challenges. A suggestion could look like this

As I see it the SAPIs we want to support are

- CLI
- phpdbg
- Apache
- FPM
- CGI

[...]

PR #4035, opened 2017-03-10
php-couchbase: PHP 5.3 is not supported anymore

avsej:
Removing it in separate PR as requested in #4026


alcaeus:
IMO, it would make sense leaving it until all other php53 formulae are removed.

There has been some discussion about this in #3686 and more recently #4005, but nothing more than that. There are tons of formulae that are no longer supported, including all of php 5.3 itself.

javian:
I would agree with that - lets keep it until we can graveyard 5.3 properly.

PR #4041, opened 2017-03-16
composer@1.2 (new formula)

arnested:

-----
A versioned composer. Version 1.2.4 -- the last version before composer
changed its date format.


javian:
Thanks for the submission @arnested . Can you please change this into @1.2 instead of @1.2.4 - this seems to be the general way to do it and would be the best in case there was a bugfix ever released for it.

javian:
Also do you have any external reference to the problem you are describing and the change of the date format ?

arnested:
Hi @javian,

I changed it to `@1.2` as suggested.

Composer changed it date format from
```
"time": "2017-01-06 14:33:56"
```
to
```
"time": "2017-01-06T14:33:56+00:00"
```
in version `1.3.0`.

If cooperating with other people who are (for some reason) stuck on an older version will lead to either inconsistent timestamps in `composer.lock` or pull requests changing a lot of lines unnecessary. This gives Homebrew users the choice to stay behind together with the people stuck on the old version.

javian:
@arnested I'm thinking we should add a caveat definition just to say that the binary/phar name is different don't you think just to be explicit about it. What do you think ?

arnested:
@javian, Makes sense. I have added a short caveats. Feel free to suggest another text for it.

javian:
@arnested looks good, lets run with that. Thank you!

PR #4044, opened 2017-03-18
php70 7.0.17

alcaeus:

-----


javian:
Thanks @alcaeus !

neutric:
Just curious: what is the deal with these "symbol not found" errors when testing xdebug? I had these in 7.0.16 (#4003) as well.

javian:
@neutric the reason is that one of the modules requires a thread safe binary which means that php is compiled with thread safety which is not compatible with the bottled non-threadsafe php modules which causes the errors.

neutric:
@javian Thanks for taking time to explain, I appreciate it!

javian:
@neutric thanks for understanding that tounge twister sentence - it was hard to read it even for me at first glance looking back at it =)

neutric:
It wasn't hard to understand, don't worry.

PR #4049, opened 2017-03-20
Replace `homebrew/versions/bison27` with `bison@2.7`

msbit:
- ☑️  Have you checked to ensure there aren't other open [Pull Requests](https://github.com/Homebrew/homebrew-php/pulls) for the same formula update/change?
- ☑️  Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)?
- ☑️  Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)?

-----

On `master`, executing `brew audit --strict php70` we get:

```
Warning: Use bison@2.7 instead of deprecated homebrew/versions/bison27
homebrew/php/php70:
  * Dependency 'homebrew/versions/bison27' was renamed; use new name 'bison@2.7'.
Error: 1 problem in 1 formula
```

This PR updates that for using the `--HEAD` build on any formula, general builds for `php53`, and removes references to `homebrew/versions` now that there are no formulae that depend on that tap.


msbit:
CI errors due to not tapping `homebrew/dupes`?

`Error: No available formula with the name "homebrew/dupes/zlib"`

javian:
@msbit its from the php..-qr packages. We shouldn't need the dupes reference anymore so that line can be remove from the formula.

msbit:
@javian `homebrew/dupes` removed in https://github.com/Homebrew/homebrew-php/pull/4059

msbit:
@javian any guidance on the build errors?

javian:
They are normal so it shouldn't be an issue and we need to fix them in other ways.

README.md
```diff
@@ -7,7 +7,6 @@ A centralized repository for PHP-related brews.
 * [Homebrew](https://github.com/Homebrew/brew)
 * Yosemite, El Capitan, and Sierra. Untested everywhere else.
 * The Homebrew `dupes` tap - `brew tap homebrew/dupes`
```
javian: You can remove the dupes reference as well.

Formula/php54-qr.rb
```diff
@@ -14,8 +14,6 @@ class Php54Qr < AbstractPhp54Extension
     sha256 "6d27fa82f3126945c3f518561bf003f0b05f5e0fd27a4cfe9eada4e6c943e098" => :yosemite
   end
 
-  depends_on "homebrew/dupes/zlib"
```
javian: As noted you need to adjust the configure line in all QR formula files.

msbit:
@javian updated the README, removing `homebrew/dupes` instructions.

javian:
Can you fix the outstanding issue as well ?

msbit:
@javian which outstanding issues are you referring to? All references to `homebrew/dupes` have been removed in both README and code.

javian:
@msbit You need to update the configure options in the qr formulas since they reference the the duped zlib module which should point to the stock osx path instead or perhaps even be left out.

msbit:
@javian ah, I see what you mean now.

I've moved `php-{53, 54, 55}-qr` to act the same as `php-56-qr`, with the zlib dir set to be `/usr`. It looks like the configure and build process was ignoring the specific `with-qr-tiff` argument and setting the linker flags for the system zlib anyway, but this is more explicit.

javian:
Thanks @msbit ! Now merged.

msbit:
👌

PR #4053, opened 2017-03-20
Point README readers to LICENSE

danielbachhuber:
Avoids maintaining two copies of the license


javian:
Thanks @danielbachhuber !

PR #4084, opened 2017-04-01
Update phpmyadmin url

BePsvPT:
After phpmyadmin 4.7.0, dependencies are managed by composer. Installing from git repo will not contains its dependencies, so we should change the url source.


javian:
Thanks @BePsvPT !

PR #4092, opened 2017-04-05
igbinary 2.0.3

BePsvPT:
---


javian:
There are a lot of dependencies to this formula - do they still work even after upgrading to this version or do they need a new revision ?

BePsvPT:
Sorry, i didn't test other formulas that depends with this formula. Is there any suggestion for this situation?

javian:
Thanks @BePsvPT ! We'll see if the dependencies fails and we can issue a revision bump for those that do.

javian:
@BePsvPT I just realised that the only thing this formula does it to provide the source code. You should update the php??-igbinary.rb files - can you do that ?

PR #4007, opened 2017-02-22
phan 0.9.0

BePsvPT:
---

Pull request opened by [homebrew-updater](https://github.com/BePsvPT/homebrew-updater) project.

PR #4114, opened 2017-04-11
phan 0.9.1

adambaratz:
Looks like this change was attempted previously in #4007. Did this by hand and tested it, works fine on my computer. I kept the previous version since there's a split in supporting PHP 7.0.x vs. 7.1.x. The test was changed from `-h` to `--help` because of a little bug where the short form exits with an error. I posted a PR on etsy/phan to fix, but might as well clean up the formula so no one's confused running `brew test phan`. I left the previous version as is.


Aliases/phan@0.9.1
```diff
@@ -0,0 +1 @@
+../Formula/phan.rb
```
javian: Can you rename this file to phan@0.9 instead so that we don't have to update the link for every version update ?

javian:
The problem in the last PR wasn't the Formula but how the dependencies are set up in the php tap.

adambaratz:
I can make that change. I don't quite understand your other comment, though. Is there something else I should do to get this merged?

Formula/phan@0.8.3.rb
```diff
@@ -0,0 +1,31 @@
+require File.expand_path("../../language/php", __FILE__)
+
+class Phan < Formula
```
javian: You need to update the class name
```
In formula file: /usr/local/Homebrew/Library/Taps/homebrew/homebrew-php/Formula/phan@0.8.3.rb
Expected to find class PhanAT083, but only found: Phan.
```

Formula/phan@0.8.3.rb
```diff
@@ -0,0 +1,31 @@
+require File.expand_path("../../language/php", __FILE__)
```
javian: Also, along the lines of the alias please name this file @0.8 instead of 0.8.3.

javian:
Lets change the reviewed items first and see if we can solve the other issue afterwards.

adambaratz:
I just updated the branch. Please let me know if there's anything else I should change. I tried to follow the [pull request instructions](http://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request.html) when putting this together. Let me know if there's anything else I should keep in mind when making changes.

javian:
@adambaratz I just realised we can't merge these formulas since they both depend on different versions of PHP. Can you take out the 0.8 version and submit a separate PR for that ?

adambaratz:
Can do. Is the idea that you want them in separate commits, or should the content of the 0.8 formula be different somehow?

javian:
@adambaratz they need to be seperate PRs, not just commits. Content wise it is ok but you don't need to include a bottle block in a new formula but it will be replaced by the bot anyway.

adambaratz:
Updated this PR so it no longer has the 0.8 content. I'll post that as a separate PR once this goes through.

javian:
Once #4129 gets merged we should be able to get this merged as well since composer is currently blocking.

javian:
@BrewTestBot test this please

javian:
Thanks @adambaratz !

PR #4129, opened 2017-04-21
composer: remove homebrew php dependency

javian:

-----

### DIFF
--- a/Formula/composer.rb
+++ b/Formula/composer.rb
@@ -40,7 +40,7 @@ class Composer < AbstractPhpPhar
     caveats = <<-EOS
     composer no longer depends on the homebrew php Formulas since the last couple of macOS releases
     contains a php version compatible with composer. If this has been part of your workflow 
-    previously then please make the appropriate changes and  brew install php71  or other appropriate
+    previously then please make the appropriate changes and `brew install php71` or other appropriate
     Homebrew PHP version.
     EOS
   end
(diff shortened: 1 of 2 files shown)


Formula/composer.rb
```diff
@@ -34,4 +35,13 @@ def phar_wrapper
       exit($return_var);
     EOS
   end
+
+  def caveats
+    caveats = <<-EOS
+    composer no longer depends on the homebrew php Formulas since the last couple of macOS releases
+    contains a php version compatible with composer. If this has been part of your workflow 
+    previously then please make the appropriate changes and  brew install php71  or other appropriate
```
adambaratz: There are two spaces between "and  brew" and "php71  or".

javian: Yeah that was intentional to separate the command from the text since I don't have any highlighting options available afaik. It could be written in a different way though.

javian: Thanks for the feedback though ^^ do you think it looks weird ?

adambaratz: I didn't realize you were doing it to separate the command. Back ticks might work for that and look more intentional. But it's not really a big deal either way.

JohnnyWalkerDesign:
Hmm. Trying to uninstall brew's php70 (so I don't have to deal with it anymore when updating php71) results in the following error:

> Error: Refusing to uninstall /usr/local/Cellar/php70/7.0.17_9
> because it is required by composer 1.4.1, which is currently installed.

Is it OK to ignore this message?

javian:
@JohnnyWalkerDesign it should be yes, if you see any issues from doing that then let us know =)

JohnnyWalkerDesign:
Seemed fine!

PR #4131, opened 2017-04-24
igbinary 2.0.4

BePsvPT:
---


BePsvPT:
@BrewTestBot test this please

javian:
There are a number of errors in Jenkins like this
```
homebrew/php/php53-memcached:
  * Dependency libmemcached should not use option with-sasl
  * macOS has been 64-bit only since 10.6 so ENV.universal_binary is deprecated.
  * macOS has been 64-bit only so build.universal? is deprecated.
```

javian:
Thanks @BePsvPT !

PR #4142, opened 2017-05-01
phan 0.8.3

adambaratz:
Now that #4114 is in, here's the formula for anyone with PHP 7.0 who wants to use the 0.8 branch of this tool.


javian:
Thanks @adambaratz !

PR #4143, opened 2017-05-01
Add support for building PHP with WebP

c4l3b:
Tested install of all supported PHP versions (5.5, 5.6, 7.0, 7.1) and ensured that the GD WebP functions existed. 

-----

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -37,7 +37,7 @@ class AbstractPhp < Formula
     depends_on "icu4c"
     depends_on "imap-uw" if build.include?("with-imap")
     depends_on "jpeg"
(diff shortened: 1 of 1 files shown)


javian:
I would prefer if these dependencies were set as  `:optional` rather than the way its done now since we need less code to accomplish the same thing. So you would write `depends_on "fluid-synth" => :optional` replacing the formula with the correct name and then adding the if statements behind it just like you have done and then you can skip line 88-90.

The php5x would then be called libvx rather than webp but I would say that's ok for now.

c4l3b:
Thank you -
Made your suggested changes & tested php5x with libvpx and php7x with webp.

javian:
Thanks @c4l3b ! Nice one

PR #4193, opened 2017-05-31
wp-cli 1.2.0

danielbachhuber:


javian:
Thanks @danielbachhuber! Can you make the following additional change to the Formula ?
```
Error: 1 problem in 1 formula
homebrew/php/wp-cli:
  * The URL http://wp-cli.org/ could use HTTPS rather than HTTP
```

danielbachhuber:
> Can you make the following additional change to the Formula ?

Updated

ilovezfs:
Thanks @danielbachhuber

PR #4226, opened 2017-06-18
Update drupalconsole to 1.0.0-rc21

gapple:

-----
Brew audit fails because the formula version is 1.0.0, but this is still an `rc` release


ilovezfs:
The audit error is caused by a version parsing issue. The version is detected from the URL as 1.0.0 before and after the URL change, so the audit is correctly triggered by the change in checksum without a change in version.

We'll need to manually override the version with
```
version "1.0.0-rc21"
```
and add
```
version_scheme 1
```
since otherwise it will consider 1.0.0-rc21 < 1.0.0.

Or we can just wait for the actual 1.0.0 to be released, and add `revision 1`, which would prevent the need to add `version_scheme 1`. Unfortunately, if we do add `version_scheme 1` it can never be removed from the formula.

A middle ground option would be to override the version to only add
```
version "1.0.0-rc21"
```
and not add version scheme, and accept that users will not be automatically upgraded from rc16 -> rc21.

javian:
@ilovezfs considering we've been doing similar updates since December then there must be a number of people who have not been upgraded over the course of time. If we added the `version_scheme` and there would be additional rc releases put out then would we have to bump the scheme number or would it then detect it properly ?

ilovezfs:
Superseded by https://github.com/Homebrew/homebrew-php/pull/4397.

PR #4323, opened 2017-07-17
php7{0-2} - New sodium.so extension for 2.x of libsodium-php

moderndeveloperllc:
The new 2.x version of libsodium-php creates a sodium.so unlike the 1.x branch, so a new Formula is needed. Also, BC in userland code is broken because the namespace has changed from \Sodium\* to the root namespace with a sodium_* prefix as [this will be a core module in 7.2](https://wiki.php.net/rfc/libsodium).

-----


moderndeveloperllc:
@BrewTestBot test this please

moderndeveloperllc:
@BrewTestBot Test me again

javian:
Hi, since the 7.2 beta has been tagged and includes this extension you can remove that file.
As for the other 2 I would prefer if we updated the existing sodium formulas for 7.0 7.1 rather than creating a new one. If we for any reason need the old 1.x version we will add versioned formulas to support that.

moderndeveloperllc:
@javian Not sure if we can just update the other two files for a couple of reasons: 

1. It is a full backwards compatibility break. User code needs to change from using `\Sodium\*` calls to `sodium_*` calls. Really shouldn't break code on just a `brew upgrade`.
2. The extension name is actually different. It goes from `libsodium` to `sodium` so the name of the file and the extension would wind up being different.
3. PHP 5.x will still use the 1.x branch.

For 7.2, there are still files for things like mcrypt, intl, yaml, etc. that have been in the core for years. What's the decision criteria? Just curious. I have no problem removing it of course.

javian:
@moderndeveloperllc its the extensions that have external dependencies that are placed in a separate formula so that the main php binary doesn't have to be rebottled if a new version is released of the dependency. As you point out though this would place the sodium extension as a separate core extension so it should have its own file for now. This might change in the future but for now we'll run with this.

Generally the formula should contain the latest version of the software. If there are backwards compatibility issues we should create a versions alias so that the extension can be installed to maintain compatibility. The pecl page has not reflected the name change https://pecl.php.net/package/libsodium so I think we should mirror that here and update the existing formulas and add the old version for 7.0/.1 .

Also, as far as I can tell yaml extension is not part of core =)

To sum it up
* Update the existing formulas with a versioned formula for lib sodium 1.0 for 7.0/7.1
* Once 7.2 beta 1 has been merged then we can submit a core extension formula for sodium similar to intl and the other core extensions

moderndeveloperllc:
@javian I don't know how to do the versioning for the formulas since the extension name and class names have to be the same. Is there an example you know of a similar extension?

moderndeveloperllc:
@javian It looks like you [pushed 7.2 beta 1](https://github.com/Homebrew/homebrew-php/commit/b83024382abaea5d5f646c28cd2699a81f72231f). Do you have any time to help me out with this? Thanks.

PR #4401, opened 2017-08-23
php72 7.2.0beta3

alcaeus:

-----


alcaeus:
Audit report:
```
Error: 2 problems in 1 formula
homebrew/php/php72:
  * Stable: version 7.2.0beta3 is redundant with version scanned from URL
  * Stable version URLs should not contain beta
```

I'm not sure what to do about the first error - should we omit `version PHP_VERSION` from built-in formulae?
As for the second error, I agree - beta should not be marked as stable. Is it possible to only have a `devel` block in a formula and no stable version until the release of 7.2?

javian:
@alcaeus hmm, yeah it looks quite new indeed. I know that not too long ago they updated the url scanning methods to better detect versions and which might originate from that. I guess you could try to remove it locally and see if it fixes the audit warning ?

For the second question I don't know off the top of my head. I would guess that a devel only section in the formula would not be allowed and we would need to put a devel section in php71 for that to work until 7.2 is properly released (this is the way it will work in the core php formula).

alcaeus:
@javian re the second point, that's quite unfortunate, as it would forbid installing PHP 7.1 and 7.2 in parallel, which for me is a common usecase. When checking for `version` I'll also see if audit complains about not having a stable section.

javian:
@alcaeus its quite the opposite. Once the devel option exists you can switch between stable and devel by executing `brew switch <version>` so devel does not replace stable and they can be installed side by side.

alcaeus:
Wow. How did I not know that?!?!?

javian:
@alcaeus I reacted in the same way when I found out not too long ago =)

javian:
@alcaeus thanks for the update!

PR #4425, opened 2017-08-29
Bump revision for php72-phalcon

alcaeus:


mignz:
Only Phalcon 3.3 works on PHP 7.2.

PR #4484, opened 2017-09-26
terminus: update tarball sha256 due to git bugfix

peff:
The hash update is discussed further in https://github.com/Homebrew/homebrew-core/issues/18044.

This was found by running the same script as was used for homebrew-core in https://github.com/Homebrew/homebrew-core/pull/18073.

/cc @ilovezfs


peff:
This was the only broken case found by my script.

SMillerDev:
Thanks @peff !

PR #4621, opened 2017-12-02
Build PHP 7.2 with libsodium by default

ngocphamm:
I don't understand the structure of brew formulae completely, but it seems that when the `depends_on "libsodium" => :recommended if name.split("::")[2].downcase.start_with?("php72")` is added, homebrew will automatically add an option `--without-libsodium`. With that said, the current check `if build.with? "sodium"` never gets to true. Changing to `unless build.without? "libsodium"` works for me locally.

-----

### DIFF
--- a/Abstract/abstract-php.rb
+++ b/Abstract/abstract-php.rb
@@ -397,7 +397,7 @@ INFO
       args << "--enable-maintainer-zts"
     end
 
-    if build.with? "sodium"
+    unless build.without? "libsodium"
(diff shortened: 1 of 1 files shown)


javian:
Thanks for your submission. The error is that the if statement doesn't check the correct name which should be `libsodium` and not `sodium` as it is now. build.with and build.without are completely interchangeable and not the cause of this issue. Since I won't have a lot of time this week I'll submit a pr #4624 and get it merged tonight and close this - thanks for the find.

PR #4750, opened 2018-02-02
Attempting to clarify caveat about installing php

dtb:
Related to #4744, trying to make it clearer what the caveat is trying to tell you. I think the key takeaway should be that you should install PHP7 (or whatever) yourself, if you need it.


javian:
Thanks @dtb !

Issue #4607, opened 2017-11-28
php71 mongoDB ext crashing php


aleksandardragojlovic:
After I install php71-mongodb my php stop working.

I have installed on clean system (High Sierra) php71 with necessary extensions. When I install mongoDB my php stop working. In system.log I see segmentation errors. When I uninstall mongoDB php ext. everything is working correctly.

➜  brew doctor
Your system is ready to brew.

➜  brew list -la
total 0
drwxrwxr-x  67 aleksandar  admin  2144 28 Nov 22:55 .
drwxr-xr-x  14 root        wheel   448 26 Nov 19:29 ..
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:13 apr
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:13 apr-util
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:38 autoconf
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:38 automake
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:14 boost
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:13 c-ares
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:43 colordiff
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:54 composer
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:43 curl
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:49 fontconfig
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:49 freetype
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:50 gd
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:37 gdbm
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:38 geoip
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:13 gettext
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 19:37 git
drwxr-xr-x   3 aleksandar  admin    96 28 Nov 11:55 httpd
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:43 icoutils
drwxr-xr-x   3 aleksandar  admin    96 26 Nov 20:13 icu4c
[...]

PR #4765, opened 2018-02-12
php{55-72}-mongodb 1.4.0

bashofmann:
- [x] Have you built your formula locally prior to submission with `brew install <formula>` (where `<formula>` is the name of the formula you're submitting)? (only php72-mongodb)
- [x] Does your submission pass `brew audit --strict --online <formula>` (after doing `brew install <formula>`)? (only php72-mongodb)

-----

### DIFF
--- a/Formula/php55-mongodb.rb
+++ b/Formula/php55-mongodb.rb
@@ -4,14 +4,9 @@ class Php55Mongodb < AbstractPhp55Extension
   init
   desc "MongoDB driver for PHP."
   homepage "https://pecl.php.net/package/mongodb"
-  url "https://pecl.php.net/get/mongodb-1.3.4.tgz"
-  sha256 "c78190115c0d51a440d66c75b6c12192f6d97873d141b34c5c2406a816fe1bb2"
+  url "https://pecl.php.net/get/mongodb-1.4.0.tgz"
+  sha256 "b970fce679b7682260eacdd1dbf6bdb895ea56e0de8a2ff74dc5af881e4d7d6a"
   head "https://github.com/mongodb/mongo-php-driver.git"
-
-  devel do
-    url "https://pecl.php.net/get/mongodb-1.4.0beta1.tgz"
-    sha256 "427185ef5a668ccb79dcdbbd853b74664cee8fe0fbb9ce241ebc5919e36c3ac7"
-  end
 
   bottle do
     sha256 "6c4613d8d81cf1f423654e01f0376dd2294bf5e36f186da5ac570eadc3d2531a" => :high_sierra
--- a/Formula/php55-mongodb.rb
+++ b/Formula/php55-mongodb.rb
@@ -14,15 +14,13 @@ class Php55Mongodb < AbstractPhp55Extension
     sha256 "33e91cfcf2dfd7a5e441614eb6c29b1600211f577ec8b54749368679b0d216ac" => :el_capitan
   end
 
-  depends_on "openssl"
-
   def install
     Dir.chdir "mongodb-#{version}" unless build.head?
 
     safe_phpize
     system "./configure", "--prefix=#{prefix}",
                           phpconfig,
-                          "--with-openssl-dir=#{Formula["openssl"].opt_prefix}"
+                          "--with-mongodb-ssl=darwin"
     system "make"
     prefix.install "modules/mongodb.so"
     write_config_file if build.with? "config-file"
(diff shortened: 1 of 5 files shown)


javian:
@BrewTestBot test this please

javian:
This would resolve #4762

jmikola:
Note that `ext-mongodb` 1.4.0 includes a new `./configure` option to utilize Secure Transport on macOS. Quoting the [1.4.0 release notes](https://github.com/mongodb/mongo-php-driver/releases/tag/1.4.0):

> This release overhauls the way the extension configures SSL libraries and adds support for LibreSSL and macOS native SSL (Secure Transport). When building from source, `configure` now supports a `--with-mongodb-ssl` option. This defaults to `auto`, which will first attempt to use OpenSSL, followed by LibreSSL, and finally Secure Transport (for macOS only, of course). Additionally, you may specify `openssl`, `libressl`, or `darwin` to force a particular library, or specify `no` to disable SSL entirely. Installing the driver via PECL will use the default `auto` behavior.

I would suggest replacing `"--with-openssl-dir=#{Formula["openssl"].opt_prefix}"` in the PHP 5.5+ formulae with `--with-mongodb-ssl=darwin`. This should address outstanding segfaults on macOS High Sierra due to BoringSSL:

 * https://github.com/Homebrew/homebrew-php/issues/4607
 * https://github.com/mongodb/mongo-php-driver/issues/681
 * https://github.com/mongodb/mongo-php-driver/issues/682

bashofmann:
@jmikola I changed the formulas to use --with-mongodb-ssl=darwin. Works fine for me with our test suite.

javian:
I think it would be better to stay with openssl since that would be the default in all other places where we use ssl. I'm assuming that it would have the same effect and keep the formula as close to what it's been previously.

jmikola:
> I think it would be better to stay with openssl since that would be the default in all other places where we use ssl. I'm assuming that it would have the same effect and keep the formula as close to what it's been previously.

I'd argue that keeping OpenSSL in place is doing a disservice to those using Apache and High Sierra, since they're inevitably going to hit the BoringSSL issue. I'm not aware if Homebrew can take options at install time (if you wanted to make `--with-mongodb-ssl` configurable), but if not that will necessitate users resorting to building the extension from source outside of Homebrew.

I read your [httpd-users mailing list post](http://markmail.org/message/ysrcf25xiluydpfg) regarding the related OpenSSL/BoringSSL crashes with SNMP (from https://github.com/Homebrew/homebrew-core/pull/16067), but it's too early to say when/if that might be addressed upstream.

I'd also point out that the [libmongoc formula](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mongo-c-driver.rb#L30) already uses Secure Transport instead of OpenSSL. If anything, I might suggest changing this PHP formula depend on the `mongo-c-driver` Homebrew package and rely on loading libmongoc/libbson as system libraries (` --with-libbson=yes --with-libmongoc=yes`) instead of using the bundled versions we ship with the PECL package. That is the same approach used for other package managers (e.g. Remi's CentOS and Red Hat packages).

javian:
I must have misunderstood the implementation then in the mongodb extension and using SecureTransport sounds like it would be the only option for now. I'm assuming the ssl implementation is isolated to the extension and has no dependencies in php at all so stacking different SSL implementations in different binaries would not end up being an issue (at least in this case) ?

As for you libmongoc option I think that sounds like a good plan moving forward and how we should implement it in the core formula.

jmikola:
> I'm assuming the ssl implementation is isolated to the extension and has no dependencies in php at all so stacking different SSL implementations in different binaries would not end up being an issue (at least in this case)?

Early versions of the driver did utilize PHP and its OpenSSL extension for our socket communication, but this was changed in 1.2.0 with [PHPC-605](https://jira.mongodb.org/browse/PHPC-605). Since 1.2.0, all socket communication and TLS utilizes libmongoc directly. The only commonality is that our extension was still linking to libssl and libcrypto, as evidenced by the BoringSSL issue.

In 1.2.x and 1.3.x, we were still using the `PHP_SETUP_OPENSSL()` macro in `config.m4` to _detect_ OpenSSL the same way that PHP itself and other dependent extensions do. In 1.4.0 we also removed that and ported over libmongoc's own [`CheckSSL.m4`](https://github.com/mongodb/mongo-php-driver/blob/1.4.0/scripts/build/autotools/CheckSSL.m4) to now detect OpenSSL, LibreSSL, and Secure Transport on our own.

Hope that clears things up.

brancengregory:
Bump

javian:
Merged now @bashofmann and thanks for all the info @jmikola ! Potentially we need to review this if/when apache httpd is fixed but this should be good for now.

